PR #11, opened 2014-09-29
Cosine and Mahalanobis distance

pradeep-pyro:
This PR implements Cosine and Mahalanobis distances.


kortschak:
At least one of the concerns raised here are no longer valid; there *is* now a dependency on mat64 in stat. What is the path forward on this PR?

btracey:
I think the general point is the same about the cosine of the angle being in floats. It's not clear to me that it belongs in stat. Is it a statistical distance? I've never seen it in that context, but I also don't know all of statistics. gonum/stat shouldn't be a dumping ground for all possible distance functions.

The Mahalanobis distance definitely feels more like a statistical distance (it's clearly related to the normal distribution). Should it actually be a method on the normal distribution instead? I wonder if mat64 should provide an InnerInv function (x' A^-1 x). It happens in mat64, and the general operation happens in the real BFGS update (though BFGS itself tracks the inverse hessian). There's a note in distmv/normal about X' A^-1 X as well (as a method on SymDense).

kortschak:
That's fair enough. Should we close this and do the necessary things in floats and mat64? (The work needed to resolve conflicts is probably more than just opening a new PR for the Mahalanobis distance code).

btracey:
Yes.

PR #20, opened 2014-12-07
Added R2 function and panic tests.

btracey:


jonlawlor:
The labels x and y are opposite to their typical use with x as predictor and y as response, so I would like you to swap them.  I'll put together some examples that I trust for the tests.

jonlawlor:
Also, there is an analogous bootstrapped version of RSquared that I ran into in http://www.iasbs.ac.ir/chemistry/chemometrics/history/4th/5.pdf called Q^2, where you blank out segments of the sample, perform the model fit, estimate the values in the missing part, and find their residuals.  I've found it particularly useful.  There may be a nice way to express models in general so that they can "automatically" get access to various goodness of fit tests.

btracey:
Do you mean the order in the function, or the definition of what the characters mean?

Bootstrap is definitely important. The way I see it is that gonum/stat implements "core" algorithms, and more advanced functions can live in various subpackages to stat. Bootstrap on its own, for example, could just return a [][]int telling you which points to sub-sample. I'm not quite sure what you're imagining, but the correct way to approach is through composition as much as possible.

jonlawlor:
I mean that the definition of the characters should be swapped.  The typical label for the response is y, not x:
in matlab:
b = regress(y,X) returns a p-by-1 vector b of coefficient estimates for a multilinear regression of the responses in y on the predictors in X. X is an n-by-p matrix of p predictors at each of n observations. y is an n-by-1 vector of observed responses.

in R:
fit <- lm(y ~ x1 + x2 + x3, data=mydata)

And because R^2 is almost always associated with a linear regression, we should stick to that nomenclature.

jonlawlor:
On second thought, in this case it would probably be clearer to label it as `y` for the response and `yhat` or `yHat` for the predicted values.

jonlawlor:
Here are a set of tests:
```go
	for i, test := range []struct {
		x   []float64
		y   []float64
		w   []float64
		ans float64
	}{
		{
			x:   []float64{-2, 0, 0, 2},
			y:   []float64{-1.76, -0.48, 0.16, 2.08},
			w:   nil,
			ans: 0.96,
		}, {
			x:   []float64{0, 2, 2, 4},
			y:   []float64{0.24, 1.52, 2.16, 4.08},
			w:   nil,
			ans: 0.96,
		}, {
			x:   []float64{-2, 0, 0, 2},
			y:   []float64{1.76, 0.48, -0.16, -2.08},
			w:   nil,
			ans: -2.88,
		},
	} {
		copyX := make([]float64, len(test.x))
		copy(copyX, test.x)
		copyY := make([]float64, len(test.y))
		copy(copyY, test.y)
		var copyW []float64
		if test.w != nil {
			copyW = make([]float64, len(test.w))
			copy(copyW, test.w)
		}
		v := RSquared(test.x, test.y, test.w)
		if !floats.Same(copyX, test.x) {
			t.Errorf("x changed for case %d", i)
		}
		if !floats.Same(copyY, test.y) {
			t.Errorf("y changed for case %d", i)
		}
		if !floats.Same(copyW, test.w) {
			t.Errorf("w changed for case %d", i)
		}
		if v != test.ans {
			t.Errorf("mismatch case %d expected: %v, found: %v", i, test.ans, v)
		}
	}
```

I fiddled with various regressions in matlab until I got one with a "nice" r squared, which is the first one.  The second one demonstrates that it RSquared(y + d, x + d) = RSquared(y, x), and the last one shows that RSquared can be negative, which I calculated manually.

kortschak:
Is this active?

btracey:
It can be. I'll incorporate @jonlawlor 's comments.

btracey:
This is already implemented.

PR #24, opened 2015-01-10
Fix Histogram implementation.

btracey:
The former behavior of Histogram did not agree with the documentation. The documentation matched
the spirit of floats.Within, so keep the documentation and fix the behavior. This change updates
the function behavior, as well as corrects the test and the example.

### DIFF
--- a/stat.go
+++ b/stat.go
@@ -413,10 +413,10 @@ func Histogram(count, dividers, x, weights []float64) []float64 {
 	if !sort.Float64sAreSorted(x) {
 		panic("histogram: x data are not sorted")
 	}
+	for i := range count {
+		count[i] = 0
+	}
 	if len(x) == 0 {
-		for i := range count {
-			count[i] = 0
-		}
 		return count
 	}
 	if x[0] < dividers[0] {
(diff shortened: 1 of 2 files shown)


btracey:
PTAL @jonlawlor

stat.go
```diff
@@ -385,7 +385,7 @@ func Hellinger(p, q []float64) float64 {
 
 // Histogram sums up the weighted number of data points in each bin.
 // The weight of data point x[i] will be placed into count[j] if
-// dividers[j-1] <= x < dividers[j]. The "span" function in the floats package can assist
+// dividers[j] <= x < dividers[j+1]. The "span" function in the floats package can assist
```
jonlawlor: Could you also add a note about what happens to the values already in count if it isn't nil?  Specifically, that it isn't zero'ed.

jonlawlor:
Otherwise LGTM.

btracey:
Oh, good catch,  it should be zeroed. Still look good when I make that
change and add a test?

jonlawlor:
SGTM

PR #25, opened 2015-01-11
Fix ExampleHistogram for new floats behavior

btracey:


btracey:
Merging because fixes test.

TBR @jonlawlor

PR #27, opened 2015-01-19
Use MulTrans

jonlawlor:
This change takes advantage of the better performance of MulTrans.  It also restricts the input weights to be non-negative.  The added benchmarks indicated that the performance was 30-40% faster when using sqrt and then MulTrans compared to the previous approach.

### DIFF
--- a/covariancematrix.go
+++ b/covariancematrix.go
@@ -7,6 +7,7 @@ package stat
 import (
 	"github.com/gonum/floats"
 	"github.com/gonum/matrix/mat64"
+	"math"
 )
 
 // CovarianceMatrix calculates a covariance matrix (also known as a
@@ -17,16 +18,17 @@ import (
 // The weights wts should have the length equal to the number of rows in
 // input data matrix x. cov should either be a square matrix with the same
 // number of columns as the input data matrix x, or nil in which case a new
-// Dense matrix will be constructed.
+// Dense matrix will be constructed.  Weights cannot be negative.
 func CovarianceMatrix(cov *mat64.Dense, x mat64.Matrix, wts []float64) *mat64.Dense {
 	// This is the matrix version of the two-pass algorithm. It doesn't use the
 	// additional floating point error correction that the Covariance function uses
 	// to reduce the impact of rounding during centering.
 
+	// TODO(jonlawlor): indicate that the resulting matrix is symmetric, and change
+	// the returned type from a *mat.Dense to a *mat.Symmetric.
+
 	r, c := x.Dims()
 
-	// TODO(jonlawlor): indicate that the resulting matrix is symmetric, which
-	// should improve performance.
 	if cov == nil {
 		cov = mat64.NewDense(c, c, nil)
 	} else if covr, covc := cov.Dims(); covr != covc || covc != c {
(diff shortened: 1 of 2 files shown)


covariancematrix.go
```diff
@@ -25,8 +26,6 @@ func CovarianceMatrix(cov *mat64.Dense, x mat64.Matrix, wts []float64) *mat64.De
 
 	r, c := x.Dims()
 
-	// TODO(jonlawlor): indicate that the resulting matrix is symmetric, which
```
btracey: Why is this TODO being removed? We probably want this to return *mat64.Symmetric when that happens.

covariancematrix.go
```diff
@@ -38,35 +37,42 @@ func CovarianceMatrix(cov *mat64.Dense, x mat64.Matrix, wts []float64) *mat64.De
...
 	var n float64
 	if wts != nil {
-		if wr := len(wts); wr != r {
-			panic(mat64.ErrShape)
-		}
 
+		// Multiply by the sqrt of the weights, so that multiplication is symmetric.
+		sqrtwts := make([]float64, r)
+		for i, w := range wts {
+			if w < 0 {
```
btracey: Should this panic, or should it just return a bunch of NaN?

covariancematrix.go
```diff
@@ -38,35 +37,42 @@ func CovarianceMatrix(cov *mat64.Dense, x mat64.Matrix, wts []float64) *mat64.De
...
+		cov.MulTrans(&xt, false, &xt, true)
+
+		// Scale by the sample size.
+		cov.Scale(1/(n-1), cov)
+		return cov
 	}
+	n = float64(r)
 
-	cov.Mul(&xt, &xc)
+	cov.MulTrans(&xt, false, &xt, true)
```
btracey: Since the non-weighted case is simpler and more common, it should probably be 
````
if weights == nil {
 // do simple stuff
}
// Complicated code

covariancematrix.go
jonlawlor: It seems like the kind of thing we would panic on, but without the check it would produce a matrix of NaNs.  The nice thing about panic is that if we expand out to allow negative weights, callers to the function would (probably) not have to change anything.  If it was all NaN, they might have some code checking for it.  I don't feel very strongly about it though.

covariancematrix.go
jonlawlor: Sure thing.

covariancematrix.go
btracey: @kortschak ?

covariancematrix.go
jonlawlor: OK, I'll put it back.  I just figured that since MulTrans was using the Symmetric route, that this was basically satisfied.  But you're right, when *mat64.Symmetric is available this function should return that type instead.

That brings up a question: do we want CovarianceMatrix to return a mat64.Matrix now, or do we want to change the type to *mat64.Symmetric later?

btracey: I think we want to change it to *mat64.Symmetric later. In either event code is going to break, as it will be common to cast the returned matrix to an explicit type. In this case, we do want to promise the explicit type of matrix returned, because it will always be some form of dense matrix. The mat64.Matrix interface is great, but we need to be careful to not overuse it.

covariancematrix.go
kortschak: Agree with @jonlawlor, though from what @btracey said previously re: negative weights, that is not the justification. NEgative weights are out of the expected range of parameters, justifying a panic.

jonlawlor:
This was from two weeks ago.  PTAL @btracey, I believe it is ready to merge.

btracey:
LGTM

jonlawlor:
Alright, I'll try out the proposed merge workflow.

Issue #28, opened 2015-01-19
Distribution Interface


jonlawlor:
It seems like the dist package could use an interface that all distributions satisfy.  Currently it only has ParameterMarshaler, but all of the distributions have:

```
CDF(x float64) float64
Entropy() float64
ExKurtosis() float64
LogProb(x float64) float64
LogSurvival(x float64) float64
Mean() float64
Median() float64
NumParameters() int
Prob(x float64) float64
Quantile(p float64) float64
Rand() float64
Skewness() float64
StdDev() float64
Survival(x float64) float64
Variance() float64
```

Although it seems like some of the other methods (such as Mode and Fit) would also be candidates for a Distribution interface.  I'd like to start a discussion on what the dist package's goals are as well.

btracey:
The signatures were designed with interfaces in mind, but I was leaning toward not having them until it was clear what methods should belong together. I would like a sampling package that has MCMC, Importance Sampling, Latin-Hypercube, etc. It's not clear how these should look because of the univariate/multivariate problem. Those aren't goals for the dist package specifically, but they bigger picture ideas.

jonlawlor:
I guess E-M and bayesian estimation would also be applications.  I'm going to flesh out some of the tests, and then I'll try to write up multivariate normal.  From there I might take a stab at some sampling.

btracey:
I've been playing with multivariate normal. Part of my motivation for the PRs on Symmetric and Triangular was to implement MVGauss. The covariance matrix should be symmetric. You need to compute the Cholesky decomposition to find probabilities, and Cholesky should return a triangular matrix. 

In any event, it's still not clear to me if it should be in dist or in a different package mvdist. I'm leaning to one package right now, but I've felt strongly in the other direction in the past. 

Here's what I have at the moment. There are lots of things wrong with it, but it's better than starting from scratch. http://play.golang.org/p/zjyKjDeR7H

btracey:
After a lot more development, I think interfaces this big are not a good idea. Instead, we should focus on what the smaller interfaces needed for specific purposes.

PR #31, opened 2015-01-28
Add NumParameters to ParameterMarshaler

jonlawlor:
The doc said, “Marshal and Unmarshal are to be used as a pair, users
should not attempt to construct the Parameter slice themselves,” but
this is incorrect, because callers have to create a []Parameter with
the appropriate length.  I’ve added NumParameters to the
ParameterMarshaler interface to make the construction possible.

Other possibilities would be to change UnmarshalParameters to take 0
inputs, and return the []Parameters after constructing it for the
caller.  We could also make NumParameters into its own NumParameterser
interface.


jonlawlor:
This is related to #28

dist/general.go
```diff
@@ -11,15 +11,15 @@ type Parameter struct {
 // ParameterMarshaler exists to support algorithms that modify parameters of arbitrary distributions.
 // Typically, users should modify distributions using the fields of the specifc
 // distribution.
-// Marshal and Unmarshal are to be used as a pair, users should not attempt to construct
```
btracey: Could we rephrase this rather than remove it? I agree the wording is weird, but the warning is useful. I would suggest something like "users should not attempt to construct the input to UnmarshalParameters themselves"

jonlawlor:
PTAL @btracey

btracey:
LGTM

kortschak:
Why not fix the behaviour by making the UnmarshalParameters method take a *[]Parameter?

jonlawlor:
I don't follow @kortschak - how would that allow the caller to construct []Parameter(s) of the appropriate size?

kortschak:
```
func foo(n int, p *[]Parameter) {
    *p = make([]Parameter, n)
    fill(*p)
}
```

This is how json etc fill nil slices. You need a way to determine the size, but that is presumable available from somewhere.

jonlawlor:
I see now.  I wasn't even at that part yet - this PR is just to make it available from somewhere.

The next set of steps will then be:

* Change the ParameterMarshaler interface to be compatible with json encoding
* Implement tests for round trip encoding / decoding
* Implement ParameterMarshaler for the remainder of the distributions that don't have it yet

kortschak:
Well, where are the parameters being unmarshaled from?

kortschak:
eek, marshaled to.

I'm not entirely clear on the bigger picture of the design here.

kortschak:
Having a stab, I think you are preparing for some kind of intermediate representation that can then be handles by other codecs. True?

Why not just have a pair of methods in the interface, `SetParameters([]Parameter)` and `Parameters() []Parameter`? I don't really see why the slice needs to be handed in to the marshaler. Allocation for something like this should not really be an issue and a return is much easier to understand than a slice to be filled or a *slice to be dereferenced. The only reason json etc take a pointer is that they don't know the type - here we do.

btracey:
Thanks for bringing this up Dan. 

The bigger picture is to have a way of abstracting over a modifiable distribution. For example, if you wanted to find the best-fit distribution to a set of points, you might use maximum likelihood estimation. You can write that individually for a specific distribution, but you can also construct an optimization algorithm that finds the best fit for an abstract distribution (then possibly trying a set of distributions). Probabilistic Graphical Models use this, for example, to find an optimum prior that fits the data.

It seems like the correct fix is to do what we do in mat64.Dense.Row, for example, and have Parameters( []Parameter) []Parameter, allocating if the input is nil.

Unfortunately, for the use case I imagine, the current implementation is not sufficiently useful. Many parameters have constraints and/or are much more effective when transformed. In a normal distribution, for example, one should optimize over log(Sigma) rather than Sigma directly.  Log(Sigma) is both unconstrained and has much more favorable numeric properties. The current method set neither presents these constraints nor provides transformations.

Another plausible use for marshaling parameters is to operate between distributions. For example, there could be a "Conjugate() []Parameter" that returns the parameters of its conjugate distribution. This, combined with UnmarshalParameters, may be a way of interoperating between dist and mvdist without having cyclic imports (Same thing could go with a Marginal that returns the parameters of a Marginal distribution). The big picture is not yet clear. It does seem clear that this is in partial conflict with the optimization goals stated above. If you want to set a normal distribution from a list, the natural representation is {mu, sigma}. If you are optimizing the parameters of the normal, you should definitely be using {mu, log(sigma)}.

kortschak:
Ahh, OK. One question and one comment.

Q: How much do you think the allocation load will be if Parameters did not take an input slice?

C: Don't call these Marshal and Unmarshal; it puts people in the mindset of serialisation rather than conversion. It's true that it is serialisation, but only as an indirect consequence of the end goal.

btracey:
A: I think that will depend on the size of the dataset, and the quality of the GC. It'll be one allocation per update, which in a small dataset could be reasonably frequently. On the other hand, if it's a large dataset, the cost of evaluating LogProbability will dwarf the allocation. If the GC becomes generational, I would imagine these are short-lived objects, so the final cost is low.

The other answer is that it seemed more in-line with gonum to allow the slice to be put in-place or to allow it to just be allocated for you (like is done in mat64)

C: Okay. Maybe these functions should instead be removed until we have a better handle on their use cases? I've used both of the things I've mentioned (optimization of distribution parameters and setting via conjugates), but I haven't yet done it in go. It's certain that to support optimization we need to answer the constraints / log issue. It's not certain to me how to best deal with the conjugate issue.

jonlawlor:
OK, then how about I remove the Marshal and Unmarshal functions, and change the tests to work without them?

I think that the distributions should be serializable, so after that how about I implement MarshalJSON and UnmarshalJSON?

kortschak:
Don't remove them. Change the signatures to match what was described above (taking the []Parameter pass through parameter as Brendan says for the Parameters method) and make them all unexported for now. You can then use them as intermediates in the json.{Marshaler,Unmarshaler} implementations. We can export them later when the needs are clearer.

jonlawlor:
Alright.  It will have to be tomorrow though.

PR #32, opened 2015-01-30
Make Marshaler internal and rename

jonlawlor:
I’ve changed `MarshalParamers` and `UnmarshalParameters` to
`getParameters` and `setParameters`, respectively.  I’ve also modified
the testConjugateUpdate function to be a table driven test, and perform
more comparisons.  It turns out that by modifying its signature, it
wasn’t necessary to use the new `parameterMutator` interface, although
that will still be useful for other serialization.

I implemented `setParameters` and `getParameters` as discussed
previously, but it would also be possible to write a single 
`Parameters([]Parameter) []Parameter` which would act as both set and
get in the same manner as `runtime.GOMAXPROCS`.

This follows up #31, PTAL @btracey, @kortschak

### DIFF
--- a/dist/exponential.go
+++ b/dist/exponential.go
@@ -123,17 +123,6 @@ func (e Exponential) LogProb(x float64) float64 {
 	return math.Log(e.Rate) - e.Rate*x
 }
 
-// MarshalParameters implements the ParameterMarshaler interface
-func (e Exponential) MarshalParameters(p []Parameter) {
-	nParam := e.NumParameters()
-	if len(p) != nParam {
-		panic("exponential: improper parameter length")
-	}
-	p[0].Name = "Rate"
-	p[0].Value = e.Rate
-	return
-}
-
 // Mean returns the mean of the probability distribution.
 func (e Exponential) Mean() float64 {
 	return 1 / e.Rate
@@ -229,8 +218,8 @@ func (e Exponential) Survival(x float64) float64 {
 	return math.Exp(-e.Rate * x)
 }
 
-// UnmarshalParameters implements the ParameterMarshaler interface
-func (e *Exponential) UnmarshalParameters(p []Parameter) {
+// setParameters modifies the parameters of the distribution.
+func (e *Exponential) setParameters(p []Parameter) {
 	if len(p) != e.NumParameters() {
 		panic("exponential: incorrect number of parameters to set")
 	}
@@ -243,4 +232,17 @@ func (e *Exponential) UnmarshalParameters(p []Parameter) {
 // Variance returns the variance of the probability distribution.
 func (e Exponential) Variance() float64 {
 	return 1 / (e.Rate * e.Rate)
+}
+
+// parameters returns the parameters of the distribution.
+func (e Exponential) parameters(p []Parameter) []Parameter {
+	nParam := e.NumParameters()
+	if p == nil {
+		p = make([]Parameter, nParam)
+	} else if len(p) != nParam {
+		panic("exponential: improper parameter length")
+	}
+	p[0].Name = "Rate"
+	p[0].Value = e.Rate
+	return p
 }
--- a/dist/general.go
+++ b/dist/general.go
@@ -5,21 +5,3 @@ type Parameter struct {
 	Name  string
 	Value float64
 }
-
-// A ParameterMarshaler is a type that can marshal itself into a slice of Paramaters
-// and unmarshal itself from the slice of parameters.
-// ParameterMarshaler exists to support algorithms that modify parameters of arbitrary distributions.
-// Typically, users should modify distributions using the fields of the specifc
-// distribution.
-// Marshal and Unmarshal are to be used as a pair, users should not attempt to construct
-// the Parameter slice themselves.
-//
-// Both MarshalParameters and  UnmarshalParameters will
-// panic if the length of the slice is not equal to the number of parameters.
-// UnmarshalParameters will panic if the names of the parameters do not match.
-// UnmarshalParameters tests names in the same order as they were created in
-// MarshalParameters.
-type ParameterMarshaler interface {
-	MarshalParameters([]Parameter)
-	UnmarshalParameters([]Parameter)
-}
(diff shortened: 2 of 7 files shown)


dist/exponential.go
```diff
@@ -229,8 +218,26 @@ func (e Exponential) Survival(x float64) float64 {
...
 
-// UnmarshalParameters implements the ParameterMarshaler interface
-func (e *Exponential) UnmarshalParameters(p []Parameter) {
+// Variance returns the variance of the probability distribution.
+func (e Exponential) Variance() float64 {
+	return 1 / (e.Rate * e.Rate)
+}
+
+// getParameters implements the parameterMutator interface
+func (e Exponential) getParameters(p []Parameter) []Parameter {
```
btracey: Shouldn't this just be "parameters"? Go generally doesn't use "Get" unless otherwise necessary.

dist/general.go
```diff
@@ -6,20 +6,17 @@ type Parameter struct {
...
 }
 
-// A ParameterMarshaler is a type that can marshal itself into a slice of Paramaters
-// and unmarshal itself from the slice of parameters.
-// ParameterMarshaler exists to support algorithms that modify parameters of arbitrary distributions.
-// Typically, users should modify distributions using the fields of the specifc
-// distribution.
-// Marshal and Unmarshal are to be used as a pair, users should not attempt to construct
+// A parameterMutator is a type that can both get and set the parameters of a distribution.
+// Typically, users should modify distributions using the fields of the specifc distribution.
```
btracey: I don't think this "users" comment is necessary now that it's an internal function. There aren't any users involved.

dist/general_test.go
```diff
@@ -64,126 +61,73 @@ func testDistributionProbs(t *testing.T, dist UniProbDist, name string, pts []un
...
-		t.Errorf("prior weight doesn't match unitary weights. First is %v, second is %v", w2, w5)
-	}
+	Rand() float64
+}
 
-	// Lastly, make sure it's okay with a bunch of random weights
-	weights := make([]float64, len(samps))
-	for i := range weights {
-		weights[i] = rand.Float64()
+func testConjugateUpdate(t *testing.T, newFittable func() ConjugateUpdater, isEqual func(ConjugateUpdater, ConjugateUpdater) bool) {
```
btracey: I don't like the second argument here, the reflection (in exponential for example) isn't necessary. We we can use getParameters here instead and compare the parameters directly.

dist/exponential.go
kortschak: Unless it collides with an existing name `parameters`. In which case add to TODO to name it `Parameters` when it is exported.

dist/general.go
kortschak: Spelling of "specific" if kept.

dist/general.go
```diff
@@ -6,20 +6,17 @@ type Parameter struct {
...
-// MarshalParameters.
-type ParameterMarshaler interface {
-	MarshalParameters([]Parameter)
-	UnmarshalParameters([]Parameter)
+// Both getParameters and setParameters will panic if the length of the slice is not equal to the
+// number of parameters, unless nil is provided to getParameters, in which case a new slice of
+// appropriate size will be allocated.
+// setParameters will panic if the names of the parameters do not match. setParameters tests names
+// in the same order as they were created in getParameters.
+type parameterMutator interface {
```
kortschak: I don't like this name, but I can't think of anything better.

kortschak:
>it would also be possible to write a single
Parameters([]Parameter) []Parameter which would act as both set and
get in the same manner as runtime.GOMAXPROCS.

That is not possible in this case, since you may be passing in a non-zero `[]Parameter` merely to save the allocation.

The reordering of source is a little confusing in the diff here (Variance moves so adds diff lines that aren't relevant AFAICS).

kortschak:
>It turns out that by modifying its signature, it
wasn’t necessary to use the new `parameterMutator` interface

Then drop it I think and add it when we need it.

dist/general_test.go
```diff
@@ -64,126 +61,73 @@ func testDistributionProbs(t *testing.T, dist UniProbDist, name string, pts []un
...
+	for i, test := range []struct {
+		samps   []float64
+		weights []float64
+	}{
+		{
```
btracey: Could you recombine the nil case and the ones case? Maybe with a "if weights == nil, test with ones". Those two answers should be the same, which is otherwise untested. I found bugs in my initial implementations because of checking that case.

dist/general.go
jonlawlor: Could call it `Parametric`, `Parameterizable`, or even split it into `ParameterGetter` and `ParameterSetter`.

jonlawlor:
I reordered it because it seemed like the style was to have unexported functions at the bottom of files.  If you want I can move it back.

Otherwise I've made the requested changes, PTAL @kortschak, @btracey.

kortschak:
Yes, please revert the order change; that can be part/all of another PR.

I still don't like that we have an interface that's only barely used. Can we drop `parameterized` and use the explicit methods in `ConjugateUpdater`. In the long run I would prefer that we use (when they are exported), `ParameterSetter` and `ParameterGetter` (still ugh, but probably the best available) as separate interfaces. Clients can define their own parameterized (without the `NumParameters`, or with, as they need).

jonlawlor:
Alright, I've moved `Variance` back and also removed the `parameterized` interface.

I think that the `parameters` and `setParameters` functions are going to be much slower than direct field access.  If we need to compare distributions, a set of hypothesis test functions would probably be faster.  In the worst case, we could use the delegate approach I was using previously in the `testConjugateUpdate` function.

kortschak:
Would you remove the comment references to parameterized since it no longer exists.

jonlawlor:
Done

jonlawlor:
I think this is ready to merge, PTAL @btracey, @kortschak

jonlawlor:
Also, on my list of things to do:

* add `CorrelationMatrix` to stat
* add `LogNormal` to dist
* add `Student` to dist (name ok?)

kortschak:
LGTM

PR #33, opened 2015-02-01
Add CorrelationMatrix

jonlawlor:
This change includes a function to calculate correlation the
correlation matrix of input data, and a function which can convert
covariance matrices to correlation matrices.

I will be adding more test cases.

### DIFF
--- a/covariancematrix.go
+++ b/covariancematrix.go
@@ -12,13 +12,13 @@ import (
 
 // CovarianceMatrix calculates a covariance matrix (also known as a
 // variance-covariance matrix) from a matrix of data, using a two-pass
-// algorithm. The matrix returned will be symmetric, square, and
-// positive-semidefinite.
+// algorithm. The matrix returned will be symmetric and square.
 //
 // The weights wts should have the length equal to the number of rows in
-// input data matrix x. cov should either be a square matrix with the same
-// number of columns as the input data matrix x, or nil in which case a new
-// Dense matrix will be constructed.  Weights cannot be negative.
+// input data matrix x. If c is nil, then a new matrix with appropriate size will
+// be constructed.  If c is not nil, it should be a square matrix with the same
+// number of columns as the input data matrix x, and it will be used as the receiver
+// for the covariance data.  Weights cannot be negative.
 func CovarianceMatrix(cov *mat64.Dense, x mat64.Matrix, wts []float64) *mat64.Dense {
 	// This is the matrix version of the two-pass algorithm. It doesn't use the
 	// additional floating point error correction that the Covariance function uses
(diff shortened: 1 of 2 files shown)


covariancematrix.go
```diff
@@ -80,3 +80,43 @@ func CovarianceMatrix(cov *mat64.Dense, x mat64.Matrix, wts []float64) *mat64.De
...
+	// TODO(jonlawlor): indicate that the resulting matrix is symmetric, and change
+	// the returned type from a *mat.Dense to a *mat.Symmetric.
+
+	// This will panic if the sizes don't match, or if wts is the wrong size.
+	corr = CovarianceMatrix(corr, x, wts)
+	Cov2Corr(corr)
+	return corr
+}
+
+// Cov2Corr converts a covariance matrix to a correlation matrix.
```
btracey: Do we need this exported? If we provide ways to compute both then we don't also need a public routine to convert them, right? Plus, this will then cause peolpe to ask where the inverse Corr2Cov function is.

jonlawlor: I made it exported because a user might want to convert between the two, and may not have the sample data used to generate the covariance.  Both R and Matlab define it (in R core, and in the Financial Toolbox in Matlab), so I figure there is a need.  I can implement the inverse Corr2Cov if you want.

kortschak: Whatever the decision, s/2/To/ please.

jonlawlor:
So, would you like me to make the conversion functions unexported?  It is up to you @btracey.

kortschak:
In the first instance make them unexported - at least until the type for the input has stabilised (TODOs on lines 106 and 130). Is that (and the return type) waiting on the symmetric type stabilisation in mat64?

jonlawlor:
Yes, that is the intent of those TODOs.  I don't think they'll last long, but I didn't think they belonged in this PR.  I'll make the conversion functions unexported.

jonlawlor:
Alright, this seems ready for merging.  PTAL @btracey, @kortschak.

covariancematrix.go
```diff
@@ -80,3 +80,71 @@ func CovarianceMatrix(cov *mat64.Dense, x mat64.Matrix, wts []float64) *mat64.De
 	cov.Scale(1/(n-1), cov)
 	return cov
 }
+
+// CorrelationMatrix calculates a correlation matrix from a matrix of data,
+// using a two-pass algorithm. The matrix returned will be symmetric, square,
+// and positive-semidefinite.
+//
+// The weights wts should have the length equal to the number of rows in
+// input data matrix x. cov should either be a square matrix with the same
```
btracey: s/cov/corr

covariancematrix.go
```diff
@@ -80,3 +80,71 @@ func CovarianceMatrix(cov *mat64.Dense, x mat64.Matrix, wts []float64) *mat64.De
...
+	return corr
+}
+
+// covToCorr converts a covariance matrix to a correlation matrix.
+func covToCorr(cov *mat64.Dense) {
+
+	// TODO(jonlawlor): use a *mat64.Symmetric as input.
+
+	r, _ := cov.Dims()
+
```
btracey: This can be much cheaper than two matrix multiplications. Just sweep over all of the rows and then over all of the columns.

covariancematrix.go
```diff
@@ -80,3 +80,71 @@ func CovarianceMatrix(cov *mat64.Dense, x mat64.Matrix, wts []float64) *mat64.De
...
+func corrToCov(corr *mat64.Dense, sigma []float64) {
+
+	// TODO(jonlawlor): use a *mat64.Symmetric as input.
+
+	r, _ := corr.Dims()
+
+	if r != len(sigma) {
+		panic(mat64.ErrShape)
+	}
+
```
btracey: Same with above, this can be much cheaper.

covariancematrix.go
jonlawlor: How about I just call it `c`?  The variable starts out as covariance, and then ends as correlation.  I made it `cov` so that the input parameters would be more descriptive.

jonlawlor:
I've renamed the variables `cov` and `corr` to `c`, added benchmarks, and changed the implementation as suggested, which resulted in a ~4x speedup for 10x10 matrices.  PTAL @btracey.

kortschak:
The "cov" name instance is still there in the doc comment. Change corr on L92 to cov and (corr := CovarianceMatrix(cov, x wts), or make covToCorr return a matrix and rewrite the last line as return covToCorr(cov)).

kortschak:
That's fine too.

LGTM after @btracey

covariancematrix.go
```diff
@@ -80,3 +80,75 @@ func CovarianceMatrix(cov *mat64.Dense, x mat64.Matrix, wts []float64) *mat64.De
 	cov.Scale(1/(n-1), cov)
 	return cov
 }
+
+// CorrelationMatrix calculates a correlation matrix from a matrix of data,
+// using a two-pass algorithm. The matrix returned will be symmetric, square,
+// and positive-semidefinite.
+//
+// The weights wts should have the length equal to the number of rows in
+// input data matrix x. c should either be a square matrix with the same
```
btracey: This should say something like 

````
If c is not nil, it must be a square matrix with the same number of columns as x, and the computed correlation matrix will be placed into c. If c is nil, a new Dense will be allocated and returned with the covariance matrix.
````
In short, add some text that the purpose of c is to be a receiver for the data.

File: covariancematrix.go [hunk 6]
```diff
@@ -80,3 +80,75 @@ func CovarianceMatrix(cov *mat64.Dense, x mat64.Matrix, wts []float64) *mat64.De
 	cov.Scale(1/(n-1), cov)
 	return cov
 }
+
+// CorrelationMatrix calculates a correlation matrix from a matrix of data,
+// using a two-pass algorithm. The matrix returned will be symmetric, square,
+// and positive-semidefinite.
```
btracey: I know that in theory covariance matrices are symmetric and positive semidefinite, but are we sure that's true in finite precision world as well? Basically, we can guarantee that the matrix is symmetric and square, so we may just want to say that.

btracey:
LGTM with documentation fixes.

covariancematrix.go
jonlawlor: The way the CovarianceMatrix operates should guarantee that it is symmetric, but I don't know about the positive semi-definite part.

jonlawlor:
The next step for me are
* add LogNormal to dist
* add Student or T distribution to dist (naming?)
* improve dist test coverage

PR #38, opened 2015-05-11
Fix DLogProbDParam and add tests

btracey:


btracey:
Apparently I couldn't do calculus when I added this. I could have sworn I checked, but I guess not. Sorry. 

PTAL @kortschak

dist/norm_test.go
```diff
@@ -83,3 +83,18 @@ func TestNormalProbs(t *testing.T) {
...
 }
+
+func TestNormDerivParam(t *testing.T) {
+	for _, test := range []*Normal{
+		&Normal{
+			Mu:    0,
+			Sigma: 1,
+		},
+		&Normal{
+			Mu:    0.32238,
```
btracey: These numbers are here because they're the values that triggered an incorrect answer in my code.

kortschak:
I'm afraid I think I'll have to pass this to @vladimir-ch.

kortschak:
LGTM for the tests. My head is not up to the calculus.

dist/general_test.go
```diff
@@ -157,3 +160,45 @@ func parametersEqual(p1, p2 []Parameter, tol float64) bool {
...
+	fdDeriv := make([]float64, d.NumParameters())
+
+	initParams := d.parameters(nil)
+	init := make([]float64, d.NumParameters())
+	for i, v := range initParams {
+		init[i] = v.Value
+	}
+	for _, v := range quantiles {
+		x := d.Quantile(v)
+		d.DLogProbDParam(x, deriv)
```
vladimir-ch: I think this line reuses parameters set in the previous for iteration inside the closure. If `d.setParameters(initParams)` is added before it, the tolerance below can be tightened to 1e-9.

Other than this, it LGTM.

btracey: Oh, nice catch. 

I don't think it's reasonable to have the generic test be so strict, but I'll fix the problem and tighten to 1e-6.

PR #39, opened 2015-05-31
Initial commit for distmv package. Adds Multivariate Gaussian and Mul…

btracey:
…tivariate Uniform distribution

### DIFF
--- a/distmv/general.go
+++ b/distmv/general.go
@@ -1,0 +1,28 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package distmv
+
+var (
+	badSizeMismatch  = "dist: size mismatch"
+	badZeroDimension = "dist: zero dimensional input"
+	badReceiver      = "dist: input slice is not nil or the correct length"
+)
+
+const logTwoPi = 1.8378770664093454835606594728112352797227949472755668
+
+// useAs gets a slice of size n. If len(x) == n, x is returned, if len(x) == 0
+// then a slice is returned of length n.
+func reuseAs(x []float64, n int) []float64 {
+	if len(x) == n {
+		return x
+	}
+	if len(x) == 0 {
+		if cap(x) >= n {
+			return x[:n]
+		}
+		return make([]float64, n)
+	}
+	panic(badReceiver)
+}
--- a/distmv/general_test.go
+++ b/distmv/general_test.go
@@ -1,0 +1,34 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package distmv
+
+import (
+	"math"
+	"testing"
+)
+
+type prober interface {
+	Prob(x []float64) float64
+	LogProb(x []float64) float64
+}
+
+type probCase struct {
+	dist    prober
+	loc     []float64
+	logProb float64
+}
+
+func testProbability(t *testing.T, cases []probCase) {
+	for _, test := range cases {
+		logProb := test.dist.LogProb(test.loc)
+		if math.Abs(logProb-test.logProb) > 1e-14 {
+			t.Errorf("LogProb mismatch: want: %v, got: %v", test.logProb, logProb)
+		}
+		prob := test.dist.Prob(test.loc)
+		if math.Abs(prob-math.Exp(test.logProb)) > 1e-14 {
+			t.Errorf("Prob mismatch: want: %v, got: %v", math.Exp(test.logProb), prob)
+		}
+	}
+}
--- a/distmv/normal.go
+++ b/distmv/normal.go
@@ -1,0 +1,126 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package distmv
+
+import (
+	"math"
+	"math/rand"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix/mat64"
+)
+
+// Normal is a multivariate normal distribution (also known as the multivariate
+// Gaussian distribution). Its pdf in k dimensions is given by
+//  (2 π)^(-k/2) |Σ|^(-1/2) exp(-1/2 (x-μ)'Σ^-1(x-μ))
+// where μ is the mean vector and Σ the covariance matrix. Σ must be symmetric
+// and positive definite. Use NewNormal to construct.
+type Normal struct {
+	mu    []float64
+	sigma *mat64.SymDense
+
+	chol       *mat64.TriDense
+	logSqrtDet float64
+	dim        int
+
+	src *rand.Rand
+}
+
+// NewNormal creates a new Normal with the given mean and covariance matrix.
+// NewNormal panics if len(mu) == 0, or if len(mu) != sigma.N. If the covariance
+// matrix is not positive-definite, the returned boolean is false.
+func NewNormal(mu []float64, sigma mat64.Symmetric, src *rand.Rand) (*Normal, bool) {
+	if len(mu) == 0 {
+		panic(badZeroDimension)
+	}
+	dim := sigma.Symmetric()
+	if dim != len(mu) {
+		panic(badSizeMismatch)
+	}
+	n := &Normal{
+		src:   src,
+		dim:   dim,
+		mu:    make([]float64, dim),
+		sigma: mat64.NewSymDense(dim, nil),
+		chol:  mat64.NewTriDense(dim, true, nil),
+	}
+	copy(n.mu, mu)
+	n.sigma.CopySym(sigma)
+	// TODO(btracey): Change this to the input Sigma, in case it is diagonal or
+	// banded.
+	ok := n.chol.Cholesky(n.sigma, true)
+	if !ok {
+		return nil, false
+	}
+	for i := 0; i < dim; i++ {
+		n.logSqrtDet += math.Log(n.chol.At(i, i))
+	}
+	return n, true
+}
+
+// Dim returns the dimension of the distribution.
+func (n *Normal) Dim() int {
+	return n.dim
+}
+
+// Entropy returns the differential entropy of the distribution.
+func (n *Normal) Entropy() float64 {
+	return float64(n.dim)/2*(1+logTwoPi) + n.logSqrtDet
+}
+
+// LogProb computes the log of the pdf of the point x.
+func (n *Normal) LogProb(x []float64) float64 {
+	dim := n.dim
+	if len(x) != dim {
+		panic(badSizeMismatch)
+	}
+	// Compute the normalization constant
+	c := -0.5*float64(dim)*logTwoPi - n.logSqrtDet
+
+	// Compute (x-mu)'Sigma^-1 (x-mu)
+	xMinusMu := make([]float64, dim)
+	floats.SubTo(xMinusMu, x, n.mu)
+	d := mat64.NewVector(dim, xMinusMu)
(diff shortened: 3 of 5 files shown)


btracey:
Not read ready for submission, as there are a few TODOs left, but this gives an idea of what I'm thinking.

PTAL @kortschak (and @jonlawlor if you'd like)

distmv/general.go
```diff
@@ -0,0 +1,21 @@
+package distmv
+
+var (
+	ErrSizeMismatch  = "dist: size mismatch"
```
kortschak: These aren't errors.

distmv/general.go
```diff
@@ -0,0 +1,21 @@
...
+	ErrZeroDimension = "dist: zero dimensional input"
+	ErrReceiver      = "dist: input slice is not nil or the correct length"
+)
+
+const logTwoPi = 1.8378770664093454835606594728112352797227949472755668
+
+// useAs gets a slice of size n. If len(x) == n, x is returned, if x == nil
+// a new slice is allocated, otherwise useAs panics
+func useAs(x []float64, n int) []float64 {
+	if len(x) == n {
```
kortschak: Do you really want this exact? In mat64 the test is `n < cap(f)` and then it's resliced to n.

distmv/general.go
```diff
@@ -0,0 +1,21 @@
...
+		return x
+	}
+	if x == nil {
```
kortschak: `len(x) == 0`?

distmv/general.go
btracey: This is how it was done in mat64. What is the appropriate style?

distmv/general.go
kortschak: Ahh, this is equivalent to reuseAs, not use.

distmv/general_test.go
```diff
@@ -0,0 +1,30 @@
+package distmv
```
kortschak: License.

distmv/normal.go
```diff
@@ -0,0 +1,126 @@
...
+
+import (
+	"math"
+	"math/rand"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix/mat64"
+)
+
+// NormalMV is a multivariate normal distribution (also known as the multivariate
```
kortschak: s/NormalMV/Normal/g

distmv/normal.go
```diff
@@ -0,0 +1,126 @@
...
+// Gaussian distribution). It's pdf in k dimensions is given by
```
kortschak: s/It's/Its/

distmv/uniform.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Uniform represents a multivariate uniform distribution
+type Uniform struct {
+	bounds []Bound
+	source *rand.Source
+}
+
+// NewUniform creates a new uniform distribution with the given bounds.
+func NewUniform(bounds []Bound, source *rand.Source) *Uniform {
```
kortschak: s/bounds/b/g
s/source/src/g

distmv/general.go
kortschak: There they are type error, these aren't. If you don't export, then you ca keep it this way.

btracey: SGTM.

distmv/general.go
btracey: Sure, I guess that's okay.

kortschak: Doing x = x[:0] is the equivalent to m.Reset().

distmv/normal.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	for i := 0; i < dim; i++ {
+		n.logSqrtDet += math.Log(n.chol.At(i, i))
+	}
+	return n, true
+}
+
+// Dim returns the dimension of the distribution.
+func (n *Normal) Dim() int {
+	return len(n.dim)
```
kortschak: `return n.dim`

distmv/normal_test.go
```diff
@@ -0,0 +1,111 @@
...
+			t.Errorf("Mean mismatch: Want: %v, got %v", test.mean, estMean)
+		}
+		estCov := stat.CovarianceMatrix(nil, samps, nil)
+		if !estCov.EqualsApprox(cov, 1e-2) {
+			t.Errorf("Cov mismatch: Want: %v, got %v", cov, estCov)
+		}
+	}
+}
+
+func expandMat(data [][]float64) []float64 {
```
kortschak: Can't we just use line breaks to do this?

distmv/uniform.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	for _, b := range bounds {
+		u.bounds[i].Min = b.Min
+		u.bounds[i].Max = b.Max
+	}
+}
+
+// Dim returns the dimension of the distribution.
+func (u *Uniform) Dim() int {
+	return len(u.min)
```
kortschak: u.min does not exist.

distmv/uniform.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// LogProb computes the log of the pdf of the point x.
+func (u *Uniform) LogProb(x []float64) float64 {
+	dim := len(u.min)
```
kortschak: u.min does not exist.

distmv/uniform.go
```diff
@@ -0,0 +1,92 @@
...
+// Prob computes the value of the probability density function at x.
+func (u *Uniform) Prob(x []float64) float64 {
+	return math.Exp(n.LogProb(x))
+}
+
+// Rand generates a random number according to the distributon.
+// If the input slice is nil, new memory is allocated, otherwise the result is stored
+// in place.
+func (u *Uniform) Rand(x []float64) []float64 {
+	x = useAs(x, n.dim)
```
kortschak: n should be u and u.dim does not exist.

distmv/uniform.go
```diff
@@ -0,0 +1,92 @@
...
+		logprob -= math.Log(b.Max - b.Min)
+	}
+	return logprob
+}
+
+// Mean returns the mean of the probability distribution at x. If the
+// input argument is nil, a new slice will be allocated, otherwise the result
+// will be put in-place into the receiver.
+func (u *Uniform) Mean(x []float64) []float64 {
+	x = useAs(x, n.dim)
```
kortschak: n should be u and u.dim does not exist.

distmv/uniform.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	u := &Uniform{
+		bounds: make([]Bound, len(bounds)),
+		source: source,
+	}
+	for _, b := range bounds {
+		u.bounds[i].Min = b.Min
+		u.bounds[i].Max = b.Max
+	}
+}
```
kortschak: `return u`.

distmv/uniform.go
```diff
@@ -0,0 +1,92 @@
...
+	return logprob
+}
+
+// Mean returns the mean of the probability distribution at x. If the
+// input argument is nil, a new slice will be allocated, otherwise the result
+// will be put in-place into the receiver.
+func (u *Uniform) Mean(x []float64) []float64 {
+	x = useAs(x, n.dim)
+	for _, b := range u.bounds {
+		x[i] = (b.Max + b.Min) / 2
```
kortschak: s/_/i/

distmv/normal.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// Entropy returns the differential entropy of the distribution.
+func (n *Normal) Entropy() float64 {
+	return float64(dim)/2*(1+logTwoPi) + n.logSqrtDet
```
kortschak: s/dim/n.dim/

distmv/normal.go
btracey: Ugh. Sorry.

distmv/general.go
btracey: Done.

distmv/general_test.go
btracey: Done.

distmv/normal.go

distmv/normal.go

distmv/uniform.go
btracey: Done.

distmv/normal_test.go
btracey: Done.

btracey:
PTAL

distmv/general_test.go
```diff
@@ -0,0 +1,34 @@
...
+	dist    prober
+	loc     []float64
+	logProb float64
+}
+
+func testProbability(t *testing.T, cases []probCase) {
+	for _, test := range cases {
+		logProb := test.dist.LogProb(test.loc)
+		if math.Abs(logProb-test.logProb) > 1e-14 {
+			t.Errorf("LogProb mismatch: Want: %v, got: %v", test.logProb, logProb)
```
kortschak: lc/Want/

and below.

kortschak:
The code LGTM, but it would be good to have someone with stronger stats also say OK. Failing that... SHIP IT!

btracey:
Are the concerns about correctness or API?

kortschak:
Correctness. I'm sure it's fine, but MVA is not a strong point of mine.

kortschak:
@Armadilloa16 would you have a look at this?

distmv/general_test.go
btracey: What is "Ic"?

btracey: oh. lowercase.

Armadilloa16:
Seems to make sense, although I would need to be more familiar with the go language and spend more time checking it thoroughly in order to be confident it is correct. From just looking it over quickly the concept seems sound, particularly as a starting point that can be expanded on.

btracey:
Thanks.

kortschak:
@btracey (Brendan), meet @Armadilloa16 (Lyron).

btracey:
Nice to meet you. Hope we'll be seeing you around here more!

Armadilloa16:
Hey nice to meet you too @btracey! I hope to be around more! I'm new to the whole software collaboration thing, but keen to learn and contribute. Was thinking of doing some stats as thats my background, maybe some classification stuff. Just a question of how much free time I can scrounge up haha. :)

jonlawlor:
That's usually the issue!

Also, I was looking at `math\rand\rand.go` and I noticed that the default source is safe for concurrent use, but `Rand.New()` is not.  It might be a good idea to document that the distribution objects safety is dependent on the Rand.Source object  - I was not expecting it.

btracey:
If you can find an appropriate place to comment, then sure, but it's not really our problem. None of our API call random numbers concurrently for this reason.

btracey:
Oh, I see. Your point is if they set the Source and then call the distribution concurrently.

kortschak:
In general types are not safe for concurrent use unless specified as such. I don't think we need to say anything more than "X is safe for concurrent use if src is safe for concurrent use." at NewX.

btracey:
Probably better as a package comment since it's a general property of dist and distmv.

kortschak:
Sure.

jonlawlor:
The tricky part is that it is safe for concurrent use, unless you specify your own source.  I can easily imagine a scenario where someone is testing out a MCMC simulation with the default Source (a problem that is embarrassingly parallelizable), then wants to "do things right" and uses a defined random seed, and ends up with something that is no longer safe.  I agree, a package comment is appropriate.

btracey:
You can set a random seed and still be safe for concurrent use (by setting rand.Seed(myint)). 

The issue with rand.Float64() is that it does not scale in parallel if you need a lot of random numbers. The way to fix that is to generate several *rand.Rand, and have a separate one in each concurrent thread. A well-written program will also be safe for concurrent use, since the point was to eliminate contention. Alternately, rand.Source is an interface, so the Int63() method could be designed for safe concurrent use.

PR #40, opened 2015-06-01
Initial commit for univariate advanced sampling package. Contains Lat…

btracey:
…inHypercube, Importance sampling, Rejection sampling, and MetropolisHastings

### DIFF
--- a/dist/interfaces.go
+++ b/dist/interfaces.go
@@ -1,0 +1,22 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package dist
+
+type LogProber interface {
+	LogProb(float64) float64
+}
+
+type Rander interface {
+	Rand() float64
+}
+
+type RandLogProber interface {
+	Rander
+	LogProber
+}
+
+type Quantiler interface {
+	Quantile(p float64) float64
+}
--- a/sample/example_burnin_test.go
+++ b/sample/example_burnin_test.go
@@ -1,0 +1,35 @@
+package sample
+
+import "github.com/gonum/stat/dist"
+
+type ProposalDist struct {
+	Sigma float64
+}
+
+func (p ProposalDist) ConditionalRand(y float64) float64 {
+	return dist.Normal{Mu: y, Sigma: p.Sigma}.Rand()
+}
+
+func (p ProposalDist) ConditionalLogProb(x, y float64) float64 {
+	return dist.Normal{Mu: y, Sigma: p.Sigma}.LogProb(x)
+}
+
+func ExampleMetropolisHastings_burnin() {
+	n := 1000    // The number of samples to generate.
+	burnin := 50 // Number of samples to ignore at the start.
+	var initial float64
+	// target is the distribution from which we would like to sample.
+	target := dist.Weibull{K: 5, Lambda: 0.5}
+	// proposal is the proposal distribution. Here, we are choosing
+	// a tight Gaussian distribution around the current location. In
+	// typical problems, if Sigma is too small, it takes a lot of samples
+	// to move around the distribution. If Sigma is too large, it can be hard
+	// to find acceptable samples.
+	proposal := ProposalDist{Sigma: 0.2}
+
+	samples := make([]float64, n+burnin)
+	MetropolisHastings(samples, initial, target, proposal, nil)
+
+	// Remove the initial samples through slicing.
+	samples = samples[burnin:]
+}
--- a/sample/example_rate_test.go
+++ b/sample/example_rate_test.go
@@ -1,0 +1,45 @@
+package sample
+
+import "github.com/gonum/stat/dist"
+
+func max(a, b int) int {
+	if a < b {
+		return b
+	}
+	return a
+}
+
+func ExampleMetropolisHastings_samplingRate() {
+	// See Burnin example for a description of these quantities.
+	n := 1000
+	burnin := 300
+	var initial float64
+	target := dist.Weibull{K: 5, Lambda: 0.5}
+	proposal := ProposalDist{Sigma: 0.2}
+
+	// Successive samples are correlated with one another through the
+	// Markov Chain defined by the proposal distribution. To get less
+	// correlated samples, one may use a sampling rate, in which only
+	// one sample from every few is accepted from the chain. This can
+	// be accomplished through a for loop.
+	rate := 50
+
+	tmp := make([]float64, max(rate, burnin))
+
+	// First deal with burnin
+	tmp = tmp[:burnin]
+	MetropolisHastings(tmp, initial, target, proposal, nil)
+	// The final sample in tmp in the final point in the chain.
+	// Use it as the new initial location.
+	initial = tmp[len(tmp)-1]
+
+	// Now, generate samples by using one every rate samples.
+	tmp = tmp[:rate]
+	samples := make([]float64, n)
+	samples[0] = initial
+	for i := 1; i < len(samples); i++ {
+		MetropolisHastings(tmp, initial, target, proposal, nil)
+		initial = tmp[len(tmp)-1]
+		samples[i] = initial
+	}
+}
--- a/sample/sample.go
+++ b/sample/sample.go
@@ -1,0 +1,179 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// package sample contains a set of advanced routines for sampling from
+// probability distributions.
+package sample
+
+import (
+	"math"
+	"math/rand"
+
+	"github.com/gonum/stat/dist"
+)
+
+var (
+	badLengthMismatch = "sample: slice length mismatch"
+)
+
+// LatinHypercube generates len(samples) samples using Latin hypercube sampling
+// from the given distribution. If src != nil, it will be used to generate
+// random numbers, otherwise rand.Float64 will be used.
+//
+// Latin hypercube sampling divides the cumulative distribution function into equally
+// spaced bins and guarantees that one sample is generated per bin. Within each bin,
+// the location is randomly sampled. The dist.UnitNormal variable can be used
+// for easy generation from the unit interval.
+func LatinHypercube(samples []float64, q dist.Quantiler, src *rand.Rand) {
+	n := len(samples)
+	var perm []int
+	var f64 func() float64
+	if src != nil {
+		f64 = src.Float64
+		perm = src.Perm(n)
+	} else {
+		f64 = rand.Float64
+		perm = rand.Perm(n)
+	}
+	for i := range samples {
+		v := f64()/float64(n) + float64(i)/float64(n)
+		samples[perm[i]] = q.Quantile(v)
+	}
+}
+
+// Importance sampling generates len(x) samples from the proposal distribution,
+// and stores the locations and importance sampling weights in place.
+//
+// Importance sampling is a variance reduction technique where samples are
+// generated from a proposal distribution, q(x), instead of the target distribution
+// p(x). This allows relatively unlikely samples in p(x) to be generated more frequently
+//
+// The importance sampling weight at x is given by p(x)/q(x). To reduce variance,
+// a good proposal distribution will bound this sampling weight. This implies the
+// support of q(x) should be at least as broad as p(x), and q(x) should be "fatter tailed"
+// than p(x).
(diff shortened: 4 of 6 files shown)


btracey:
The plan is to also have a "samplemv" package with multivariate versions of these routines

Couple of API questions:
1) LatinHypercube sampling. Right now, there is no randomization done of the sample locations (they are always returned in CDF order). In the multivariate version, every dimension but the first will have to be randomized (typically using rand.Perm). On the one hand, it seems better to have all the dimensions randomized, but on the other hand, it changes the univariate algorithm from O(n) to O(n log(n)) with no real benefit (any user algorithm that assumes LH samples are in random order is wrong).

2) I'm a little uneasy about the fact that some of the target distributions can actually be proportional to the probability rather than the true probability. This is asking users to implement a method that is called LogProb but is really PropLogProb. It's pretty ugly if we don't force that overloading though. Either we have PropLogProb, and then provide helper interfaces to convert real LogProb into PropLogProb, or instead of having the target be an interface, we instead make it a function (func(float64)float64). I think the function  both obscures the meaning and opens up the danger that users present the PDF rather than log(PDF).

3) The tests I have are pretty nice examples of using the methods. I would have made them as Examples, but I'm not sure how to do that given that they are randomized algorithms, and as far as I know, rand.Rand() is not guaranteed to give the same random numbers on all platforms and compilers.

dist/interfaces.go
```diff
@@ -0,0 +1,22 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package dist
+
+type LogProber interface {
+	LogProb(float64) float64
+}
+
+type Rander interface {
```
kortschak: Remove this from general_test.go

btracey: Ah, thanks. I was just looking into why it failed.

sample/sample.go
```diff
@@ -0,0 +1,174 @@
...
+// samples according to the distribution specified by target by using the Markov
+// chain implicitly defined by the proposal distribution. At each
+// iteration, a proposal point is generated randomly from the current location.
+// This proposal point is accepted with probability
+//  p = min(1, (target(new) * proposal(current|new)) / (target(current) * proposal(new|current)))
+// If the new location is accepted, it is stored into samples and becomes the
+// new current location. If it is rejected, the current location remains and
+// is stored into samples. Thus, a location is stored into samples at every iteration.
+//
+// The samples in Metrpolis Hastings are correlated with one another through the
```
kortschak: s/Metrpolis/Metropolis/

sample/sample.go
```diff
@@ -0,0 +1,174 @@
...
+// on the early samples, and so typically, the first sapmles generated by the chain.
+// are ignored. This is known as "burn-in", and can be accomplished with slicing
+//  nSamples := 1000
+//  burnin := 50
+//  samples := make([]float64, nSamples+burnin)
+//  MetropolisHastings(x, intial, target, proposal, src)
+//  x = x[burnin:]
+// The exact length of the burn-in will depend on the sampling and the target
+// distribution.
+func MetropolisHastings(samples []float64, initial float64, target dist.LogProber, proposal MHProposal, src *rand.Rand) {
```
kortschak: Do you think it's worthwhile having a sampling rate?

btracey: Do you mean like the "trim" in Matlab? I don't think so. First of all, it's controversial. Some say you should, some say you shouldn't. It's more code to implement, it creates a more complicated function signature, and there are more failure rates (if trim < 0). Secondly, I don't think it's a good idea. If you are interested in an expected value (or something), you should keep all of your samples as they are information about the function. Trim is only useful when  trying to de-correlate samples for some other purpose. If that's the case, the user can trim themselves
````
trimmed := make([]float64, len(samples) / trim)
for i := 0; i < len(samples) / trim; i++ {
    trimmed[i] = samples[i*trim]
}
````
If you think it's worthwhile to comment that segment, I can. The controversy over trimming seemed too much for the function comment.

**Comment by kortschak** [comment]
1) Can you case switch for the univariate case and not randomise then?

3) If you use a Rander rather than rand.Rand, you can make a PRNG that is the same on all platforms. The one Rob Pike uses is very simple:

```
var seed = ^uint32(0)

func Rand() float64 {
	seed += seed
	seed ^= 1
	if int32(seed) < 0 {
		seed ^= 0x88888eef
	}
	return float64(seed%10000) / 10000
}
```

File: sample/sample.go (hunk 3 above)
Comment (kortschak) [inline sample/sample.go]: What a bizarre name for that. Yes, I guess you're right. The reason I was suggesting it is exactly for when you are wanting to de-correlate; if you are wanting to do that and you are making large numbers of samples, you may be over allocating by a couple of orders of magnitude depending on the application.

Comment (btracey) [inline sample/sample.go]: I hadn't considered the allocation, which could make it reasonable. Even in that case though, the trim rate is very hard to set properly. The only ways I've seen to set it that aren't WAGs are to keep all of the samples for a while and do some statistics.

In the really large sample case, it's only a couple of lines to implement trim.
````
samples := make([]float64, nSamples)
tmp := make([]float64, rate)
for i := 0; i  < nSamples; i++ {
    MetropolisHastings(tmp, initial, target, proposal, src)
    initial = tmp[len(tmp)-1]
    samples[i] = initial
}
````
So, it depends on what we want to make easy.

Comment (btracey) [inline sample/sample.go]: The interaction between burnoff and rate is tricky. Burnoff is super easy with slicing, but if the samples were generated with a sampling rate then slicing is actually cutting off a large number of samples.

**Comment by btracey** [comment]
1) If the decision is to not scramble here, then in the multivariate case it would also be unscrambled in the first dimension always. It's easy to implement as such. The question is just if it's better to return scrambled locations.

3) Would the function be implemented in the test package somewhere? I don't want to have that code in the example itself.

File: sample/sample.go (hunk 3 above)
Comment (kortschak) [inline sample/sample.go]: I'm happy with your snippet. We could add that as an example for how to do sub-sampling, rather than providing the code directly.

**Comment by kortschak** [comment]
1) @vladimir-ch?

3) You can use an implementation that's not shown in the example.

**Comment by vladimir-ch** [comment]
I'd like to help but statistics and sampling are not my strong points. I'll be watching from afar and help with what I can.

**Comment by btracey** [comment]
It's not a technical question that's being asked in 1). Latin Hypercube samples points from a distribution in a special way. For the univariate case, one generate them O(n), but the samples are in a very specific order. Alternately, they can be scrambled in O(n log(n)) time. The multivariate case must use some scrambling (so will always be O(n log(n))), and the first dimension can either be scrambled or not. Would you have preferences as a novice user?

**Comment by kortschak** [comment]
BTW Fisher-Yates is O(n). Where is the O(nlogn) coming from? Am I missing something?

File: sample/sample.go (hunk 3 above)
Comment (btracey) [inline sample/sample.go]: I added the documentation with the snippet.

File: sample/sample.go [hunk 4]
```diff
@@ -0,0 +1,186 @@
...
+//  samples := make([]float64, nSamples+burnin)
+//  MetropolisHastings(x, intial, target, proposal, src)
+//  x = x[burnin:]
+// The exact length of the burn-in will depend on the sampling and the target
+// distribution.
+//
+// Many choose to have a sampling "rate" where a number of samples
+// are ignored in between each kept sample. This helps decorrelate
+// the samples from one another, but also reduces the number of available samples.
+// A sampling rate can be implemented with a for loop as follows
```
kortschak: New paragraph here. Does this warrant an example test?

btracey: You're thinking "A sampling rate can be implemented with successive calls to MetropolisHastings." and then have the actual loop be an example?

btracey:
The O(n log(n)) comes from using a poor algorithm.

sample/sample.go
kortschak: Ignore the new paragraph comment. What you have here is good. But yes, should the snippet be and example test? I'm not fussed either way really.

kortschak:
Here, have this one.

```
func shuffle(f []float64) {
	for i := range f {
		r := i + rand.Intn(len(f)-i)
		f[r], f[i] = f[i], f[r]
	}
}
```

vladimir-ch:
Oh, I missed the beginning of the 1)2)3) discussion. Now it makes stightly more sense, but still ... Perhaps if I understood what "scrambled" means here. So, without actually knowing what I am talking about, it seems to me that the first dimension in the multivariate case should not have a special treatment just because it happens to be the first.

btracey:
Thanks for the algorithm Dan. When you said it was actually O(n), I looked at the rand.Perm implementation (basically identical to what you just posted) and discovered you were right. A long time ago I had learned the n log(n) algorithm from Matlab and had just assumed that was the best way to do it. Thanks for the correction.

kortschak:
I'm really surprised that MATLAB does that; FY is well established.

sample/sample.go
btracey: Okay, I made them examples. However, I am using the CondNorm as the proposal distribution. I don't want to make it part of the package, but can I include the source code in the example somehow for them both to use? Or do I need to have a comment with the code?

kortschak: Yes, you make all the dependencies that are needed and to be shown as part of a single file. See the godoc for pkg/testing:

>The entire test file is presented as the example when it contains a single example function, at least one other function, type, variable, or constant declaration, and no test or benchmark functions.

sample/sample_test.go
```diff
@@ -0,0 +1,124 @@
...
+				if p < float64(i)/float64(nSamples) || p > float64(i+1)/float64(nSamples) {
+					t.Errorf("probability out of bounds")
+				}
+			}
+		}
+	}
+}
+
+func ExampleMetropolisHastings_burnin() {
+	nSamples := 1000
```
kortschak: s/nSamples/n/g (I wouldn't mind if this happened throughout).

sample/sample_test.go
```diff
@@ -0,0 +1,124 @@
...
+	burnin := 50
+	var initial float64
+	target := dist.Weibull{K: 5, Lambda: 0.5}
+	proposal := CondNorm{Sigma: 0.2}
```
kortschak: Probably have a comment explaining what this is. Remember this is a documentation example.

sample/sample_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	samples := make([]float64, nSamples+burnin)
+	MetropolisHastings(samples, initial, target, proposal, nil)
+	samples = samples[burnin:]
+}
+
+func ExampleMetropolisHastings_samplingRate() {
+	nSamples := 1000
+	trim := 50
```
kortschak: s/trim/rate/g (I think trim is a horrible name for this - I'd never heard it called this before today - and it's at odds with the name of the example).

Again comments through here to explain a bit of what is going on.

btracey: Oops. I agree that trim is a bad name, but it's habit (got it right in the title).

btracey:
commentary added. PTAL

sample/example_rate_test.go
```diff
@@ -0,0 +1,45 @@
...
+	// Successive samples are correlated with one another through the
+	// Markov Chain defined by the proposal distribution. To get less
+	// correlated samples, one may use a sampling rate, in which only
+	// one sample from every few is accepted from the chain. This can
+	// be accomplished through a for loop.
+	rate := 50
+
+	tmp := make([]float64, max(rate, burnin))
+
+	// First deal with burnin
```
kortschak: Full stop.

sample/example_rate_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+func max(a, b int) int {
+	if a < b {
+		return b
+	}
+	return a
+}
+
+func ExampleMetropolisHastings_samplingRate() {
+	// See Burnin example for a description of these quantities
```
kortschak: Full stop.

sample/example_burnin_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (p ProposalDist) ConditionalLogProb(x, y float64) float64 {
+	return dist.Normal{Mu: y, Sigma: p.Sigma}.LogProb(x)
+}
+
+func ExampleMetropolisHastings_burnin() {
+	n := 1000    // The number of samples to generate.
+	burnin := 50 // Number of samples to ignore at the start.
+	var initial float64
+	// target is the distribution from which we would like to sample
```
kortschak: Full stop.

Sorry - I keep forgetting to comment on the PR not the commit.

kortschak:
LGTM after punctuation and given the cheapness of scrmalbnig, we may as well do that for LatinHypercube.

btracey:
Huh. I don't like internal comments to be sentences (unless it's long) but the Go style guide says otherwise. Will fix.

PR #43, opened 2015-06-03
stat/sample: add weighted sampler

kortschak:
@btracey 

PTAL when you have time.

I can add sample with replacement at a later stage as method Choose([]int) which would fill the []int with n samples.

### DIFF
--- a/sample/weighted.go
+++ b/sample/weighted.go
@@ -1,0 +1,137 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this code is governed by a BSD-style
+// license that can be found in the LICENSE file
+
+package sample
+
+import (
+	"math/rand"
+
+	"github.com/gonum/floats"
+)
+
+// Weighted provides sampling without replacement from a collection of items with
+// non-uniform probability.
+type Weighted struct {
+	weights []float64
+	// heap is a weight heap.
+	//
+	// It keeps a heap-organised sum of remaining
+	// index weights that are available to be taken
+	// from.
+	//
+	// Each element holds the sum of weights for
+	// the corresponding index, plus the sum of
+	// of its children's weights; the children
+	// of an element i can be found at positions
+	// 2*(i+1)-1 and 2*(i+1). The root of the
+	// weight heap is at element 0.
+	//
+	// See comments in container/heap for an
+	// explanation of the layout of a heap.
+	heap []float64
+	src  *rand.Rand
+}
+
+// NewWeighted returns a Weighted for the weights w. If src is nil, rand.Rand is
+// used as the random source.
+//
+// Note that sampling from weights with a high variance or overall low absolute
+// value sum may result in problems with numerical stability.
+func NewWeighted(w []float64, src *rand.Rand) Weighted {
+	s := Weighted{
+		weights: make([]float64, len(w)),
+		heap:    make([]float64, len(w)),
+	}
+	s.ReweightAll(w)
+	return s
+}
+
+// Len returns the number of items held by the Weighted, including items
+// already taken.
+func (s Weighted) Len() int { return len(s.weights) }
+
+// Take returns an index from the Weighted with probability proportional
+// to the weight of the item. The weight of the item is then set to zero.
+// Take returns false if there are no items remaining.
+func (s Weighted) Take() (idx int, ok bool) {
+	const small = 1e-12
+	if floats.EqualWithinAbsOrRel(s.heap[0], 0, small, small) {
+		return -1, false
+	}
+
+	var r float64
+	if s.src == nil {
+		r = s.heap[0] * rand.Float64()
+	} else {
+		r = s.heap[0] * s.src.Float64()
+	}
+	i := 1
+	last := -1
+	left := len(s.weights)
+	for {
+		if r -= s.weights[i-1]; r <= 0 {
+			break // Fall within item i-1.
+		}
+		i <<= 1 // Move to left child.
+		if d := s.heap[i-1]; r > d {
+			r -= d
+			// If enough r to pass left child
+			// move to right child state will
+			// be caught at break above.
+			i++
+		}
+		if i == last || left < 0 {
+			// No progression.
+			return -1, false
+		}
+		last = i
+		left--
+	}
+
+	w, idx := s.weights[i-1], i-1
+
+	s.weights[i-1] = 0
+	for i > 0 {
+		s.heap[i-1] -= w
+		// The following condition is necessary to
+		// handle floating point error. If we see
+		// a heap value below zero, we know we need
+		// to rebuild it.
+		if s.heap[i-1] < 0 {
+			s.reset()
+			return idx, true
+		}
+		i >>= 1
+	}
+
+	return idx, true
+}
+
+// Reweight sets the weight of item idx to w.
+func (s Weighted) Reweight(idx int, w float64) {
+	w, s.weights[idx] = s.weights[idx]-w, w
+	idx++
+	for idx > 0 {
+		s.heap[idx-1] -= w
+		idx >>= 1
+	}
+}
+
+// ReweightAll sets the weight of all items in the Weighted. ReweightAll
+// panics if len(w) != s.Len.
+func (s Weighted) ReweightAll(w []float64) {
+	if len(w) != s.Len() {
+		panic("floats: length of the slices do not match")
+	}
+	copy(s.weights, w)
+	s.reset()
+}
+
+func (s Weighted) reset() {
+	copy(s.heap, s.weights)
+	for i := len(s.heap) - 1; i > 0; i-- {
+		// Sometimes 1-based counting makes sense.
+		s.heap[((i+1)>>1)-1] += s.heap[i]
+	}
+}
--- a/sample/weighted_test.go
+++ b/sample/weighted_test.go
@@ -1,0 +1,267 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this code is governed by a BSD-style
+// license that can be found in the LICENSE file
+
+package sample
+
+import (
+	"flag"
+	"math/rand"
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/gonum/floats"
+)
+
+var prob = flag.Bool("prob", false, "enables probabilistic testing of the random weighted sampler")
+
+const sigChi2 = 16.92 // p = 0.05 df = 9
+
+var (
+	newExp = func() []float64 {
+		return []float64{1 << 0, 1 << 1, 1 << 2, 1 << 3, 1 << 4, 1 << 5, 1 << 6, 1 << 7, 1 << 8, 1 << 9}
+	}
+	exp = newExp()
+
+	obt = []float64{973, 1937, 3898, 7897, 15769, 31284, 62176, 125408, 250295, 500363}
+)
+
+func newTestWeighted() Weighted {
+	weights := make([]float64, len(obt))
+	for i := range weights {
+		weights[i] = float64(int(1) << uint(i))
+	}
+	return NewWeighted(weights, nil)
+}
+
+func TestWeightedUnseeded(t *testing.T) {
+	rand.Seed(0)
+
+	want := Weighted{
+		weights: []float64{1 << 0, 1 << 1, 1 << 2, 1 << 3, 1 << 4, 1 << 5, 1 << 6, 1 << 7, 1 << 8, 1 << 9},
+		heap: []float64{
+			exp[0] + exp[1] + exp[3] + exp[4] + exp[7] + exp[8] + exp[9] + exp[2] + exp[5] + exp[6],
+			exp[1] + exp[3] + exp[4] + exp[7] + exp[8] + exp[9],
+			exp[2] + exp[5] + exp[6],
+			exp[3] + exp[7] + exp[8],
+			exp[4] + exp[9],
+			exp[5],
+			exp[6],
+			exp[7],
+			exp[8],
+			exp[9],
+		},
+	}
+
+	ts := newTestWeighted()
+	if !reflect.DeepEqual(ts, want) {
+		t.Fatalf("unexpected new Weighted value:\ngot: %#v\nwant:%#v", ts, want)
+	}
+
+	f := make([]float64, len(obt))
+	for i := 0; i < 1e6; i++ {
+		item, ok := newTestWeighted().Take()
+		if !ok {
+			t.Fatal("Weighted unexpectedly empty")
+		}
+		f[item]++
+	}
+
+	exp := newExp()
+	fac := floats.Sum(f) / floats.Sum(exp)
+	for i := range f {
+		exp[i] *= fac
+	}
+
+	if !reflect.DeepEqual(f, obt) {
+		t.Fatalf("unexpected selection:\ngot: %#v\nwant:%#v", f, obt)
+	}
+
+	// Check that this is within statistical expectations - we know this is true for this set.
+	X := chi2(f, exp)
+	if X >= sigChi2 {
+		t.Errorf("H₀: d(Sample) = d(Expect), H₁: d(S) ≠ d(Expect). df = %d, p = 0.05, X² threshold = %.2f, X² = %f", len(f)-1, sigChi2, X)
+	}
+}
+
+func TestWeightedTimeSeeded(t *testing.T) {
+	if !*prob {
+		t.Skip("probabilistic testing not requested")
+	}
+	t.Log("Note: This test is stochastic and is expected to fail with probability ≈ 0.05.")
+
+	rand.Seed(time.Now().Unix())
+
+	f := make([]float64, len(obt))
+	for i := 0; i < 1e6; i++ {
+		item, ok := newTestWeighted().Take()
+		if !ok {
+			t.Fatal("Weighted unexpectedly empty")
+		}
+		f[item]++
+	}
+
+	exp := newExp()
+	fac := floats.Sum(f) / floats.Sum(exp)
+	for i := range f {
+		exp[i] *= fac
+	}
+
+	// Check that our obtained values are within statistical expectations for p = 0.05.
+	// This will not be true approximately 1 in 20 tests.
+	X := chi2(f, exp)
+	if X >= sigChi2 {
+		t.Errorf("H₀: d(Sample) = d(Expect), H₁: d(S) ≠ d(Expect). df = %d, p = 0.05, X² threshold = %.2f, X² = %f", len(f)-1, sigChi2, X)
+	}
+}
+
+func TestWeightZero(t *testing.T) {
+	rand.Seed(0)
+
+	want := Weighted{
+		weights: []float64{1 << 0, 1 << 1, 1 << 2, 1 << 3, 1 << 4, 1 << 5, 0, 1 << 7, 1 << 8, 1 << 9},
+		heap: []float64{
+			exp[0] + exp[1] + exp[3] + exp[4] + exp[7] + exp[8] + exp[9] + exp[2] + exp[5],
+			exp[1] + exp[3] + exp[4] + exp[7] + exp[8] + exp[9],
+			exp[2] + exp[5],
+			exp[3] + exp[7] + exp[8],
+			exp[4] + exp[9],
+			exp[5],
+			0,
+			exp[7],
+			exp[8],
+			exp[9],
+		},
+	}
+
+	ts := newTestWeighted()
+	ts.Reweight(6, 0)
+	if !reflect.DeepEqual(ts, want) {
+		t.Fatalf("unexpected new Weighted value:\ngot: %#v\nwant:%#v", ts, want)
+	}
+
+	f := make([]float64, len(obt))
+	for i := 0; i < 1e6; i++ {
+		ts := newTestWeighted()
+		ts.Reweight(6, 0)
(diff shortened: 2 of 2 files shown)


sample/weighted.go
```diff
@@ -0,0 +1,104 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this code is governed by a BSD-style
+// license that can be found in the LICENSE file
+
+package sample
+
+import "math/rand"
+
+// Weighted is a collection of items that can be selected with weighted probabilities
```
btracey: I think this is tricky enough that it's better for the comment to express intent rather than just being a descriptive, something along the lines of

"Weighted is a type for sampling without replacement from a collection of items with weighted probabilites"

sample/weighted.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	s.reset()
+	return s
+}
+
+func (s Weighted) reset() {
+	copy(s.totals, s.weights)
+	for i := len(s.totals) - 1; i > 0; i-- {
+		// Sometimes 1-based counting makes sense.
+		s.totals[((i+1)>>1)-1] += s.totals[i]
```
btracey: I know I should be better at bit shifting, but this for loop is very hard to read. 

Seems much simpler for reset() to just be
````
floats.CumSum(totals, weights)
````
(then it doesn't even need a method)

Sampling is then (loosely)
````
r := s.totals[len(totals) - 1] * rand.Float64()
var idx int
for i, v := range totals {
   if r < v {
        idx = i 
       break
   }
}
w := totals[idx]
for i := idx; i < len(totals); i++ {
    totals[i] -= w
}
````
I may be missing something, but my version seems way simpler.

File: sample/weighted.go (hunk 1 above)
Comment (kortschak) [inline sample/weighted.go]: Funny, that's almost what it said before. I'll change it.

File: sample/weighted.go (hunk 2 above)
Comment (kortschak) [inline sample/weighted.go]: Time complexity; one is O(m + nlogm), the other is O(nm) where m is size of the collection and n is the number of samples.

File: sample/weighted.go [hunk 3]
```diff
@@ -0,0 +1,104 @@
...
+// Weighted is a collection of items that can be selected with weighted probabilities
+// without replacement.
+type Weighted struct {
+	indexes []int
+	weights []float64
+	totals  []float64
+	src     *rand.Rand
+}
+
+// NewWeighted returns a Weighted for the indexes in idx weighted by w. Elements of
```
btracey: I don't like the indexes argument. This type is for sampling from a weighted set of objects whose weights are represented by the w slice. This is fundamentally a cross between Intn() and Perm(), and like those methods should just return values between 0 and len(weight). The added semantics of being elements from an index slice are not relevant to the generality of the type. Why is it a []int as opposed to a []float64, []string or []optimize.Method? It seems to me that adding semantics for a particular kind of set are best used through embedding.

````
type EdgeSampler struct {
    sample.Weighted
    Ids []int
}

func (e EdgeSampler) Take() (id int, ok bool){
   idx, ok := e.Weighted.Take()
   if !ok {
       return -1, ok
   }
   return e.Ids[idx], true 
}

File: sample/weighted.go (hunk 2 above)
Comment (kortschak) [inline sample/weighted.go]: BTW, the complexity is why the comment exists. In a 1-based indexing system it looks like:

```
	copy(s.totals, s.weights)
	for i := len(s.totals) - 1; i > 0; i-- {
		s.totals[i>>1] += s.totals[i]
	}
```

Comment (btracey) [inline sample/weighted.go]: Okay, that's a good reason. 

There needs to be a description of the contents in totals. It's not the obvious thing (the cumulative probability).

File: sample/weighted.go (hunk 3 above)
Comment (kortschak) [inline sample/weighted.go]: That's good. I'll do that.

File: sample/weighted.go (hunk 2 above)
Comment (btracey) [inline sample/weighted.go]: Does the code look any nicer if it goes the other way? Totals [end] = sum

Comment (kortschak) [inline sample/weighted.go]: It's not a linear cumulative sum it's a heap sum. It can't run the opposite direction.

File: sample/weighted.go [hunk 4]
```diff
@@ -0,0 +1,98 @@
...
+
+package sample
+
+import "math/rand"
+
+// Weighted provides sampling without replacement from a collection of items with
+// weighted probabilites.
+type Weighted struct {
+	weights []float64
+	heap    []float64
```
btracey: I like the name update. I still think it deserves an extended comment about the data structure. In general Go is not tricky with data structures, but here we are being tricky. It's for a good cause, but since it's more complicated than it appears it deserves a comment. If there's a good description somewhere in the standard library, it's fine to link it, otherwise I'd appreciate seeing something along the lines of:

````
heap accumulates the probabilities using an in-memory binary heap to allow sampling in log(n) time rather than linear.

Each element, i, in the heap contains the weight of element i plus the sum of the weights of the nodes below it. The root of the heap (index zero) thus contains the full some of the weights. The heap is traversed by [indexing method to travel to the childen and indexing method to travel to the parent]
````

File: sample/weighted.go [hunk 5]
```diff
@@ -0,0 +1,98 @@
...
+	src     *rand.Rand
+}
+
+// NewWeighted returns a Weighted for the weights w. Elements of w are modified
+// during sampling, so a copy should be passed if the user needs to keep the values.
+// If src is nil, rand.Rand is used as the random source.
+func NewWeighted(w []float64, src *rand.Rand) Weighted {
```
btracey: I think NewWeighted should allocate and copy instead of maintaining a reference. Everything is much easier to think about when data isn't shared, and the ReweightAll method allows the Weighted to be continually reused to limit garbage.

kortschak: I'd prefer not to do this, but I will think about it.

sample/weighted.go
kortschak: The docs for container heap are pretty close.

sample/weighted.go
```diff
@@ -0,0 +1,98 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this code is governed by a BSD-style
+// license that can be found in the LICENSE file
+
+package sample
+
+import "math/rand"
+
+// Weighted provides sampling without replacement from a collection of items with
+// weighted probabilites.
```
kortschak: I just thought, this is wrong - they are not probabilities, they are weights. How would you like me to phrase that?

sample/weighted.go
```diff
@@ -0,0 +1,98 @@
...
+		// Sometimes 1-based counting makes sense.
+		s.heap[((i+1)>>1)-1] += s.heap[i]
+	}
+}
+
+// Len returns the number of items held by the Weighted, including items
+// already taken.
+func (s Weighted) Len() int { return len(s.weights) }
+
+// Take returns an index from the Weighted based on the Weighted weighting, the
```
btracey: Take returns an index from the Weighted with probability proportional to the weight of the item. The weight of the item is then set to zero. Take returns false if there are no items remaining.

sample/weighted.go
```diff
@@ -0,0 +1,98 @@
...
+	s.weights[i-1] = 0
+	for i > 0 {
+		s.heap[i-1] -= w
+		i >>= 1
+	}
+
+	return idx, true
+}
+
+// Reweight alters the weight of item i in the Weighted.
```
btracey: Reweight sets the weight of item i to w.

sample/weighted_test.go
```diff
@@ -0,0 +1,236 @@
...
+		weights[i] = float64(int(1) << uint(i))
+	}
+	return NewWeighted(weights, nil)
+}
+
+func TestWeightedUnseeded(t *testing.T) {
+	rand.Seed(0)
+
+	want := Weighted{
+		weights: []float64{1 << 0, 1 << 1, 1 << 2, 1 << 3, 1 << 4, 1 << 5, 1 << 6, 1 << 7, 1 << 8, 1 << 9},
```
btracey: Shouldn't this use newExp?

sample/weighted_test.go
```diff
@@ -0,0 +1,236 @@
...
+	f := make([]float64, len(obt))
+	for i := 0; i < 1e6; i++ {
+		item, ok := newTestWeighted().Take()
+		if !ok {
+			t.Fatal("Weighted unexpectedly empty")
+		}
+		f[item]++
+	}
+
+	exp := newExp()
```
btracey: I don't understand the purpose of this segment. Isn't it easier just to test
````
for i, v := range f {
    if v != 1{
       // Wrong number of items found error
   }
}

File: sample/weighted_test.go [hunk 11]
```diff
@@ -0,0 +1,236 @@
...
+	fac := floats.Sum(f) / floats.Sum(exp)
+	for i := range f {
+		exp[i] *= fac
+	}
+
+	if !reflect.DeepEqual(f, obt) {
+		t.Fatalf("unexpected selection:\ngot: %#v\nwant:%#v", f, obt)
+	}
+
```
btracey: We should try and take one more item from Weighted and make sure it fails.

sample/weighted_test.go
btracey: Sorry, I'm wrong. I missed the i < 1e6

sample/weighted_test.go
btracey: This is wrong too.

sample/weighted_test.go
```diff
@@ -0,0 +1,236 @@
...
+			exp[8],
+			exp[9],
+		},
+	}
+
+	ts := newTestWeighted()
+	if !reflect.DeepEqual(ts, want) {
+		t.Fatalf("unexpected new Weighted value:\ngot: %#v\nwant:%#v", ts, want)
+	}
+
```
btracey: Could we do a test of taking all of the items out? Get a new weighted, take each item out, make sure they're all there, and make sure that taking one more item out fails.

sample/weighted.go
btracey: It's not wrong, it's just an overloading of the word "weight". It is a weighted sampling, but the weights is proportional, not equal to the weight.

How about, "of items with non-uniform probability"

sample/weighted_test.go
kortschak: I'm adding tests for large random samples. I found an error - we assume things will hit zero, but they may not because of float rounding. Fixed that.

sample/weighted_test.go
kortschak: You, I'm doing that.

Finding interesting float problems that make the scale of what we can offer limited. I'm not sure how to deal with that.

sample/weighted.go
btracey: I do have a preference toward my suggestion. 

The type and its methods do not exist to modify the values in w, it exists to provide random integers, and currently has a side effect of modifying w. It's more more cognitively simple for the user to think "NewWeighted gives me a sandbox for sampling without replacement" than it is to have to worry about the state of the slice. Furthermore, at present the shared state isn't consistent. If the weights are set with ReweightAll, the slice is free to modify.  At much as possible, we should be limiting "long-range" state.

sample/weighted.go
kortschak: SGTM

sample/weighted.go
kortschak: Nice.

sample/weighted_test.go
kortschak: I want it to be explicit.

sample/weighted.go
btracey: It's especially weird to have the shared state in this case because the heap memory needs to be allocated anyway, so it's not like it changes the structure from "no allocation" to "allocation".

That said, if it is really important to save the memory for this data structure, I would be supportive of a function that explicitly asks for the shared state, something like

````
// NewWeightedShared creates a Weighted structure where the memory is provided. The data will be modified as the functions are used.
// TODO(btracey): better comment.
NewWeightedShared(weights, heap []float64)
````

**Comment by kortschak** [comment]
PTAL

We have a problem. Where the number of items is large we can have significant impact from floating point error. This in turn can result in either a bounds error or a non-halting loop. I'm not sure how to deal with this in the general case. One approach could be to add a `taken int` field that is a count of non-zero weights that is decremented when an item is taken and perform our test against that instead of `heap[0]`.

File: sample/weighted.go (hunk 5 above)
Comment (kortschak) [inline sample/weighted.go]: I'll change it.

File: sample/weighted_test.go (hunk 12 above)
Comment (btracey) [inline sample/weighted_test.go]: Is "You" supposed to be "Yes"?

Comment (kortschak) [inline sample/weighted_test.go]: Indeed.

**Comment by btracey** [comment]
Thinking.

**Comment by btracey** [comment]
The floating point error can be significant in terms of the algorithm functioning, but small in terms of absolute value, right?

One thing to check is if any of the heap lengths are < 0 after subtraction. If they are, then re-compute the heap. We can also keep a count, as you mention. These should fix the bulk of the problems. As long as the floating point skew is small in value, it shouldn't affect the sampling probability too much.

**Comment by kortschak** [comment]
The first does the job, nice work. The rebuild is rarely invoked - it should be approximately zero times in reasonable usage where a small sample of the total is being taken each time. If it becomes a problem we can add the counter; then we need to put all the methods on a pointer.

**Comment by jonlawlor** [comment]
If it helps the R source (in C) for the equivalent is at https://github.com/wch/r-source/blob/0c43709dfb69053b8669223db04a24b33d2b9724/src/main/random.c#L396

The sample may be smaller because the weights are implicit in the records of the dataframe, so there is no question of termination.  Seems like there isn't much call to do enormous samplings without replacement?

File: sample/weighted.go [hunk 13]
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/gonum/floats"
+)
+
+// Weighted provides sampling without replacement from a collection of items with
+// non-uniform probability.
+type Weighted struct {
+	weights []float64
+	// heap is a weight heap. See comments
+	// in contained/heap for an explanation.
```
btracey: container/heap

sample/weighted.go
```diff
@@ -0,0 +1,118 @@
...
+	"math/rand"
+
+	"github.com/gonum/floats"
+)
+
+// Weighted provides sampling without replacement from a collection of items with
+// non-uniform probability.
+type Weighted struct {
+	weights []float64
+	// heap is a weight heap. See comments
```
btracey: I don't think this is sufficiently descriptive. Sorry. 

container/heap says "A heap is a tree with the property that each node is the minimum-valued node in its subtree". That's misleading for a number of reasons. 1) Our sign is flipped, ours is higher (okay, whatever). 2) Both of the examples in container heap are for keeping a sorted list of values. Our heap does not maintain a sorted list of weights, nodes contain sums of subsets of weights. 3) The public documentation for container/heap does not say anything about traversing, which is needed for reading this code. At the least it should point to the implementation (see code for up/down in container heap). 4) Related to 2, the code in container/heap does not look like this code. It makes calls to Less and Swap, but there are no operations that look like Less and Swap in the code to reset. 

It's okay to use a pointer to container/heap to describe what a heap is, but should be a description of what the data in this heap actually is.

sample/weighted.go
```diff
@@ -0,0 +1,118 @@
...
+		i <<= 1 // move to left child
+		if d := s.heap[i-1]; r > d {
+			r -= d
+			// if enough r to pass left child
+			// move to right child
+			// state will be caught at break above
+			i++
+		}
+		if i == last {
+			// No progression.
```
btracey: What does this case mean? I'm surprised this doesn't result it returning an error.

kortschak: Good point. I'll make this return -1, false and add a count down from len(s.weights) since this doesn't catch cycles of length > 1.

sample/weighted.go
kortschak: 1) I'm not sure what you mean by this.

2) Is actually not correct; it does keep a sorted list of weights - admittedly not the weights given to the type, but the totaled weights.

3) I guess this is true, but I think it is generally unerstood that a heap is really just a linear representation of a tree.

4) That is because the container/heap mutate by deleting elements and needs to reattain the heap invariants by reordering, here we mutate weight and reattain invariants by adjusting.

Notwithstanding all these, I will expand the explanation.

kortschak:
PTAL

kortschak:
*ping*

sample/weighted.go
```diff
@@ -0,0 +1,133 @@
...
+// non-uniform probability.
+type Weighted struct {
+	weights []float64
+	// heap is a weight heap.
+	//
+	// It keeps a heap-organised sum of remaining
+	// index weights that are available to be taken
+	// from.
+	//
+	// Each elements holds the sum of weights for
```
btracey: Each element

sample/weighted.go
```diff
@@ -0,0 +1,133 @@
...
+	// the corresponding index, plus the sum of
+	// of its children's weights; the children
```
btracey: childrens'

kortschak: No.

sample/weighted.go
kortschak: Thanks.

sample/weighted.go
```diff
@@ -0,0 +1,133 @@
...
+	"math/rand"
+
+	"github.com/gonum/floats"
+)
+
+// Weighted provides sampling without replacement from a collection of items with
+// non-uniform probability.
+type Weighted struct {
+	weights []float64
+	// heap is a weight heap.
```
btracey: Thanks. This comment is exactly what I wanted.

sample/weighted.go
btracey: Yep, you're right, sorry.

kortschak: :)

sample/weighted.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// Len returns the number of items held by the Weighted, including items
+// already taken.
+func (s Weighted) Len() int { return len(s.weights) }
+
+// Take returns an index from the Weighted with probability proportional
+// to the weight of the item. The weight of the item is then set to zero.
+// Take returns false if there are no items remaining.
+func (s Weighted) Take() (idx int, ok bool) {
```
btracey: This "EqualWithinAbsOrRel" makes me think of two things:
1) Maybe we should have a comment on NewWeighted that is something like:
````
For reasons of numeric stability, Weighted cannot operate on very small floating point values. If floating point values are close to the precision of float64, then they should instead be renormalized to sum to roughly 1
````

2) Instead of the above, is it worth keeping track of the maximum initial weight? Then we can do EqualWithinRel on that value, and it should have better scaling over the float64 range.

btracey:
LGTM except for that last issue.

sample/weighted.go
kortschak: What would you prefer? I'm easy either way (I like neither very much, but I was thinking similar things a couple of days ago).

btracey: This is where my knowledge/intuition about floating point is insufficient, but I would imagine the latter would work better for a wide range of values. If the initial values are all O(1e4) or something, this AbsOrRel won't be the correct test. Similarly if they're all really tiny. It also feels less surprisng. If I give it 1e-13, 2e-13, and 2e-12, it feels like it should work. The latter case will also break down if the scales have a huge variance (differ by a few orders of magnitude), but at that point I (as a user) would figure that it's floating point error.

Make sure that the "max weight" is also updated during Reweight (outside reset)

kortschak: On reflection, I think the first option is better. If we take the second, we really need to throw away the Reweight method or keep careful track of weights, because the user can arbitrarily re-weight items, potentially invalidating our assumptions that the maximum held is actually a maximum. Just saying "Don't be silly." seems reasonable.

btracey: Maybe not outside reset? I don't know. Floating point is crazy.

btracey: That sounds fine. Maybe say something simpler than I proposed, like "For accuracy reasons, it's best if the weights have a magnitude around 1."

kortschak: Part of the issue is that the user really should not expect to be able to sample all the items from the sampler. It doesn't really make sense to do so in most cases (I guess if you want a weighted permutation is might). I think just a note that numerical stability may be an issue when there are large numbers of items or when there is a significant variance in the weighting should be enough.

Why do you think that the AbsOrRel won't work for numbers in O(1e4)? The worst case, AFAICS, is that we may not select all of the items over a complete set of Take iterations. We at least guarantee to terminate.

kortschak: Around 1 by what margin? In my original use of this type I was selecting graph edges based on weight, with weights ranging from 1e2 - 1e5.

btracey: I agree with the simple comment (a la signficant variance), although it's also true when the absolute value of the weights are tiny.

It's not that AbsOrRel won't work for 1e4, it's that the actual threshold value of 1e-12 may not be correct, maybe it should be 1e-10 instead. 

I'm thinking about this through the lens of what I encoutered in optimize. I wanted to avoid a large cause of linesearch failure by just saying that small variations in the function value (1e-14 or so) aren't relevant. The problem is that for any reasonable threshold one wants to define, there is always a function whose floating point noise is higher, particularly for concurrent functions. The ramifications were too complicated in optimize, so we just gave up. 

Here is different because we are only doing subtraction, but that was the motivation for the initial comment. Why 1e-12? Why not 1e-6 or 1e-14? Because it's a reasonable default if the weights are reasonable, which means roughly O(1) and not O(1e-10) or O(1e10)

kortschak:
PTAL

btracey:
LGTM

PR #45, opened 2015-06-11
Add Rander and LogProber to distmv

btracey:


distmv/interface.go
```diff
@@ -0,0 +1,13 @@
+package distmv
```
btracey: Copyright

kortschak:
Some background?

btracey:
I was/am playing around with a possible API for optimize.Global (and the test method needs this). It's across repositories, so it can't go in as one change anyway. If you'd like me to hold off until I have working code, I can do that too.

kortschak:
Having a reason in the commit message or a link to an issue would be good. I can't really review it unless I know what it needs to do.

btracey:
Okay. I'll keep it local for now then.

PR #46, opened 2015-08-03
Update stat for the new API of MulTrans and MulVec

btracey:


kortschak:
I just did the same for one of mine. It is very nice to be able to say things this way.

LGTM

btracey:
I agree. I'm very excited about this new API.

PR #49, opened 2015-09-12
Fix Normal to new Cholesky api

btracey:


btracey:
Needs https://github.com/gonum/matrix/pull/207

distmv/normal.go
```diff
@@ -21,7 +21,8 @@ type Normal struct {
 	mu    []float64
 	sigma *mat64.SymDense
 
-	chol       *mat64.TriDense
+	chol       *mat64.Cholesky
+	l          *mat64.TriDense
```
kortschak: I think this could be called lower rather than l.

distmv/normal.go
```diff
@@ -21,7 +21,8 @@ type Normal struct {
 	mu    []float64
 	sigma *mat64.SymDense
```
kortschak: sigma can be deleted since it's not used anywhere and make chol and (lower) concrete rather than pointer types and lose the &{} lines in NewNormal.

distmv/normal.go

distmv/normal.go

kortschak:
LGTM

PR #51, opened 2015-09-20
stat: avoid overloaded word

kortschak:
The word "alternate" is not universally correct English here (correct in North America and some parts of Australian speech). Use the word that is universally unambiguous.

@jonlawlor @btracey Please take a look.

### DIFF
--- a/covariancematrix_test.go
+++ b/covariancematrix_test.go
@@ -61,7 +61,7 @@ func TestCovarianceMatrix(t *testing.T) {
 			copy(w, test.weights)
 		}
 		c := CovarianceMatrix(nil, test.data, test.weights)
-		if !c.Equals(test.ans) {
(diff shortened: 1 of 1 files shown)


btracey:
Lgtm

In NA English, "alternate", meaning to change back and forth, has a distinct pronunciation from "alternate", meaning alternative. Not particularly relevant, but interesting.

kortschak:
Really. We have a differentiation between the verb and the adjective: (from google which matches what we have here /ˈɔːltəneɪt,ˈɒl-/ for the verb and /ɔːlˈtəːnət,ɒl-/ for the adjective (which matches the way you use it for the meaning we don't use - so it still looks overloaded to me).

This is a general principle in English, e.g. record (stressed first syllable - noun) and record (unstressed first syllable - verb)

PR #52, opened 2015-09-20
stat: reflect change in mat64 API for equality testing

kortschak:
@btracey Please take a look.


btracey:
Lgtm, though overlaps with the alternative name change commit

PR #60, opened 2015-10-26
stat: update matrix error imports

kortschak:
Fixes build.

@btracey Please take a look.

### DIFF
--- a/covariancematrix.go
+++ b/covariancematrix.go
@@ -8,6 +8,7 @@ import (
 	"math"
 
 	"github.com/gonum/floats"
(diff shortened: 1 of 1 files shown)


covariancematrix.go
```diff
@@ -8,6 +8,7 @@ import (
 	"math"
 
 	"github.com/gonum/floats"
+	"github.com/gonum/matrix"
```
btracey: It's so weird to see this. I'm used to it being a mistake.

btracey:
LGTM

PR #61, opened 2015-11-10
Add ConditionNormal

btracey:


btracey:
I'm not a huge fan of the name, but it seemed like we may want to have a "func Condition() Distribution" style interface that I didn't want to corrupt with the true return type.

btracey:
Needs https://github.com/gonum/matrix/pull/297 and https://github.com/gonum/matrix/pull/296

distmv/normal.go
```diff
@@ -118,3 +118,99 @@ func (n *Normal) Rand(x []float64) []float64 {
...
+// where the elements of x_2 have been observed with values v. The dimension order
+// is preserved during conditioning, so if the value of dimension 1 is observed,
+// the returned normal represents dimensions {0, 2, ...} of the original Normal
+// distribution.
+//
+// ConditionalNormal returns {nil, false} if there is a failure during the update.
+// Mathematically this is impossible, but can occur with finite precision arithmetic.
+func (n *Normal) ConditionalNormal(observed []int, values []float64, src *rand.Rand) (*Normal, bool) {
+	if len(observed) != len(values) {
+		panic("norm: input slice length mismatch")
```
kortschak: "normal:... ?

distmv/normal.go
```diff
@@ -118,3 +118,99 @@ func (n *Normal) Rand(x []float64) []float64 {
 	floats.Add(x, n.mu)
 	return x
 }
+
+// ConditionNormal returns the Normal distribution that is the receiver conditioned
```
vladimir-ch: `ConditionalNormal`

distmv/normal.go
```diff
@@ -118,3 +118,99 @@ func (n *Normal) Rand(x []float64) []float64 {
...
+//
+// ConditionalNormal returns {nil, false} if there is a failure during the update.
+// Mathematically this is impossible, but can occur with finite precision arithmetic.
+func (n *Normal) ConditionalNormal(observed []int, values []float64, src *rand.Rand) (*Normal, bool) {
+	if len(observed) != len(values) {
+		panic("norm: input slice length mismatch")
+	}
+	// Recompute original sigma
+	sigma := mat64.NewSymDense(n.Dim(), nil)
+	sigma.FromCholesky(&n.chol)
```
kortschak: Any reason not to retain a copy of the sigma handed to NewNormal?

kortschak:
"ConditionedNormal"?

distmv/normal.go
btracey: The main reason is that it's double the memory storage (for something that's already O(n^2)). I would imagine that Conditional and Marginal (where Sigma is also needed) are relatively uncommon operations, so I wasn't sure if normal users should pay the extra cost.  It will also be necssary if we add a Cov method, which may be much more common. Not clear what the right tradeoff is.

kortschak: The other option is to have a sigma field that is nil but is populated here. The basis for this being that if you are doing it once you might be wanting it again.

btracey:
Maybe just ConditionNormal? The general interface as 
````
// A Conditioner returns a new distribution which is the current distribution conditioned on the provided evidence
type Conditioner interface {
      Condition(dims []int, values []float64) Distribution
}
````

File: distmv/normal.go (hunk 3 above)
Comment (btracey) [inline distmv/normal.go]: And also add a mutex to the type? Otherwise the methods stop being thread safe and there is no other reason they shouldn't be.

Comment (kortschak) [inline distmv/normal.go]: It would be an RWMutex if you want to add it, but I don't see the need.

Comment (kortschak) [inline distmv/normal.go]: Actually, a sync.Once would be what is wanted.

File: distmv/normal.go (hunk 1 above)

File: distmv/normal.go (hunk 2 above)
Comment (btracey) [inline distmv/normal.go]: Fixed by making it Condition.

File: distmv/normal.go (hunk 3 above)
Comment (btracey) [inline distmv/normal.go]: Yes. Nice.

File: distmv/normal.go [hunk 4]
```diff
@@ -118,3 +121,109 @@ func (n *Normal) Rand(x []float64) []float64 {
...
+			v := sigma11.At(i, j)
+			sigma11.SetSym(i, j, v-tmp4.At(i, j))
+		}
+	}
+	return NewNormal(mu1, &sigma11, src)
+}
+
+// sigma computes and stores the covariance matrix of the distribution, or
+// returns it if it has already been computed.
+func (n *Normal) getSigma() *mat64.SymDense {
```
kortschak: The alternative to this is the note that Condition (and Cov) recreate the sigma and so if used in those cases is not thread safe.

distmv/normal.go
```diff
@@ -118,3 +121,109 @@ func (n *Normal) Rand(x []float64) []float64 {
...
+	n.once.Do(func() {
+		sigma := mat64.NewSymDense(n.Dim(), nil)
```
kortschak:
```
n.sigma = ...
n.sigma.From...
```

Why is it returned?

btracey: > Why is it returned?

If you read from n.sigma directly it's a data race. A future PR will implement Marginal, which also needs n.sigma. This puts it all in one place.

kortschak: What is it racing against? I don't see that here. sync.Once prevents that.

kortschak: Ah, OK. Yes.

kortschak: > Do is intended for initialization that must be run exactly once. Since f is niladic, it may be necessary to use a function literal to capture the arguments to a function to be invoked by Do:

```
config.once.Do(func() { config.init(filename) })
```

>Because *no call to Do returns until the one call to f returns*, if f causes Do to be called, it will deadlock.

This means that if you have reached a use of n.sigma later, it must be non-nil. If this weren't true, your return approach would not work either.

btracey: I don't understand. In ConditionNormal, we cannot just do
````
sigma := n.sigma
````
because that would race. We need:
a)
`````
n.computeSigma()
sigma = n.sigma
```
b)
````
sigma := n.getSigma()
````

You're right if we reach a call to n.sigma later, but we have to guarantee we reach the call to Once in the first place.

Comment (kortschak) [inline distmv/normal.go]: When you call n.getSigma(), you guarantee that by the time you have returned n.sigma holds the reconstructed sigma, so a use of n.sigma later (replace the sigma uses with n.sigma uses - I don't see why these are needed) are OK.

Comment (btracey) [inline distmv/normal.go]: Sure.

File: distmv/normal.go [hunk 6]
```diff
@@ -118,3 +121,107 @@ func (n *Normal) Rand(x []float64) []float64 {
...
+// distribution.
+//
+// ConditionalNormal returns {nil, false} if there is a failure during the update.
+// Mathematically this is impossible, but can occur with finite precision arithmetic.
+func (n *Normal) ConditionNormal(observed []int, values []float64, src *rand.Rand) (*Normal, bool) {
+	if len(observed) != len(values) {
+		panic("normal: input slice length mismatch")
+	}
+
+	n.setSigma()
```
kortschak: I was going to suggest a name change. This is good.

distmv/normal.go
```diff
@@ -118,3 +121,107 @@ func (n *Normal) Rand(x []float64) []float64 {
...
+	for i := 0; i < len(unobserved); i++ {
+		for j := i; j < len(unobserved); j++ {
+			v := sigma11.At(i, j)
+			sigma11.SetSym(i, j, v-tmp4.At(i, j))
+		}
+	}
+	return NewNormal(mu1, &sigma11, src)
+}
+
+// sigma computes and stores the covariance matrix of the distribution, or
```
kortschak: setSigma

distmv/normal.go
```diff
@@ -18,7 +19,9 @@ import (
 // where μ is the mean vector and Σ the covariance matrix. Σ must be symmetric
 // and positive definite. Use NewNormal to construct.
 type Normal struct {
-	mu []float64
+	mu    []float64
+	sigma *mat64.SymDense // only stored if needed
+	once  sync.Once
```
kortschak: Move the once ubove sigma and add a new line between that pair of line and mu.

distmv/normal.go
```diff
@@ -118,3 +121,109 @@ func (n *Normal) Rand(x []float64) []float64 {
 	floats.Add(x, n.mu)
 	return x
 }
+
+// ConditionNormal returns the Normal distribution that is the receiver conditioned
+// on the input evidence. The returned multivariate normal has dimension
+// n - len(dims), where n is the dimension of the original receiver. The updated
```
kortschak: What is dims?

distmv/normal.go
```diff
@@ -118,3 +121,109 @@ func (n *Normal) Rand(x []float64) []float64 {
...
+// mean and covariance are
+//  mu = mu_1 + sigma_{2,1}^T * sigma_{2,2}^-1 (v - mu_2)
+//  sigma = sigma_{1,1} - sigma_{2,1}^T * sigma_{2,2}^-1 * sigma_{2,1}
+// where the elements of x_2 have been observed with values v. The dimension order
+// is preserved during conditioning, so if the value of dimension 1 is observed,
+// the returned normal represents dimensions {0, 2, ...} of the original Normal
+// distribution.
+//
+// ConditionalNormal returns {nil, false} if there is a failure during the update.
```
kortschak: ConditionNormal

distmv/normal.go

distmv/normal.go
btracey: Arg, sorry.

distmv/normal.go

distmv/normal.go

kortschak:
LGTM but wait for @vladimir-ch.

distmv/normal.go
```diff
@@ -118,3 +122,107 @@ func (n *Normal) Rand(x []float64) []float64 {
...
 	return x
 }
+
+// ConditionNormal returns the Normal distribution that is the receiver conditioned
+// on the input evidence. The returned multivariate normal has dimension
+// n - len(observed), where n is the dimension of the original receiver. The updated
+// mean and covariance are
+//  mu = mu_1 + sigma_{2,1}^T * sigma_{2,2}^-1 (v - mu_2)
+//  sigma = sigma_{1,1} - sigma_{2,1}^T * sigma_{2,2}^-1 * sigma_{2,1}
+// where the elements of x_2 have been observed with values v. The dimension order
```
vladimir-ch: Maybe the notation is very standard in statistics but the comment is confusing for me without looking at Wikipedia. I like the comments to be as self-contained as reasonable which in this case is difficult but could you please at least write something about the subscripts and how they relate to observed?

distmv/normal.go
```diff
@@ -118,3 +122,107 @@ func (n *Normal) Rand(x []float64) []float64 {
...
+		for j := i; j < len(unobserved); j++ {
+			v := sigma11.At(i, j)
+			sigma11.SetSym(i, j, v-tmp4.At(i, j))
+		}
+	}
+	return NewNormal(mu1, &sigma11, src)
+}
+
+// setSigma computes and stores the covariance matrix of the distribution, or
+// returns it if it has already been computed.
```
vladimir-ch: setSigma doesn't return anything

vladimir-ch:
LGTM after addressing the documentation comments

btracey:
Okay, but gonum/matrix/#296 needs to go in first.

distmv/normal.go

distmv/normal.go

PR #64, opened 2015-11-11
Add MarginalNormal and CovarianceMatrix to normal

btracey:


btracey:
Wow, this diff looks a lot bigger than it is. What happened was adding MarginalNormal, CovarianceMatrix, and I moved ConditionalNormal to preserve alphabetical ordering.

kortschak:
Why do we need `CovarianceMatrix`? Presumably the sigma is available from during construction - is it for cases where the user has called `ConditionNormal`? If so, can't they do `ss.SubsetSym(sigma, vars)`? If it's just for convenience, I'm OK with that, but it would be nice to know.

btracey:
One reason is that you could be given the Normal (or a list of normal) rather than constructing it yourself and so may not have the Sigma. If you have computed MarginalNormal you can use the call you mention, but that is not the correct update for ConditionNormal. To find Sigma you'd have to do the whole update rule yourself. The final reason is that CovarianceMatrix is a more general operation that could easily have a reasonable interface with it (just like StdDev can be computed for many different distributions).

kortschak:
LGTM but please wait for @vladimir-ch.

distmv/normal.go
```diff
@@ -221,6 +156,106 @@ func (n *Normal) ConditionNormal(observed []int, values []float64, src *rand.Ran
...
+	floats.SubTo(xMinusMu, x, n.mu)
+	d := mat64.NewVector(dim, xMinusMu)
+	tmp := make([]float64, dim)
+	tmpVec := mat64.NewVector(dim, tmp)
+	tmpVec.SolveCholeskyVec(&n.chol, d)
+	return c - 0.5*floats.Dot(tmp, xMinusMu)
+}
+
+// MarginalNormal returns the marginal distribution of the given input variables.
+// That is, MarginalNorm returns
```
vladimir-ch: MarginalNormal

vladimir-ch:
I'm wondering if `ConditionNormal` shouldn't be called `ConditionalNormal` after all for consistency with `MarginalNormal`.

LGTM

btracey:
I think no, but I don't feel strongly. Marginal keeps the variables that
are input (it returns the marginal distribution of those variables).
Condition removes the input variables, it conditions them out

Issue #63, opened 2015-11-11
dist: Implement categorical distribution


btracey:
This will have a lot of overlap with sample/Weighted

kortschak:
@btracey this is done?

PR #66, opened 2015-11-12
Add Categorical distribution and tests

btracey:


btracey:
Fixes #63

dist/categorical.go
```diff
@@ -0,0 +1,176 @@
...
+	src *rand.Rand
+}
+
+// NewCategorical constructs a new categorical distribution where the probability
+// that x equals i is proportional to w[i].
+func NewCategorical(w []float64, src *rand.Rand) Categorical {
+	c := Categorical{
+		weights: make([]float64, len(w)),
+		heap:    make([]float64, len(w)),
+	}
```
vladimir-ch: src is not set

dist/categorical.go
```diff
@@ -0,0 +1,176 @@
...
+	return cdf / c.heap[0]
+}
+
+// Entropy returns the entropy of the distribution.
+func (c Categorical) Entropy() float64 {
+	var ent float64
+	for _, w := range c.weights {
+		p := w / c.heap[0]
+		if p == 0 {
```
vladimir-ch: `if w == 0` and move one line up?

dist/categorical.go
```diff
@@ -0,0 +1,176 @@
...
+	last := -1
+	left := len(c.weights)
+	for {
+		if r -= c.weights[i-1]; r <= 0 {
+			break // Fall within item i-1.
+		}
+		i <<= 1 // Move to left child.
+		if d := c.heap[i-1]; r > d {
+			r -= d
+			// If enough r to pass left child
```
vladimir-ch: I don't understand this.

dist/categorical.go
```diff
@@ -0,0 +1,176 @@
...
+// Reweight sets the weight of item idx to w.
+func (c Categorical) Reweight(idx int, w float64) {
+	if w < 0 {
+		panic("categorical: negative weight")
+	}
+	w, c.weights[idx] = c.weights[idx]-w, w
+	idx++
+	for idx > 0 {
+		c.heap[idx-1] -= w
+		idx >>= 1
```
vladimir-ch: `/= 2`?

dist/categorical.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	copy(c.weights, w)
+	c.reset()
+}
+
+func (c Categorical) reset() {
+	copy(c.heap, c.weights)
+	for i := len(c.heap) - 1; i > 0; i-- {
+		// Sometimes 1-based counting makes sense.
+		c.heap[((i+1)>>1)-1] += c.heap[i]
```
vladimir-ch: `/2`?

dist/categorical.go
btracey: Thanks.

dist/categorical.go
btracey: Done.

dist/categorical.go
btracey: It's saying that we can move to the right child, and if we're supposed to stop there then we will stop at the start of the for loop (so we don't need to do anything special). @kortschak ?

dist/categorical.go
btracey: I think it's standard when moving in heaps to do bit shifts like this rather than division.

dist/categorical.go
btracey: Same as above.

dist/categorical.go
kortschak: Yes. Probably clearer with comma at the end of this line.

kortschak:
I'm troubled by the extensive code duplication this brings, but I can't see a sensible way to reduce that code sharing because of the need to be able to see the weight and heap states for the distribution statistics methods here.

Is it the case that this is always wanting a sample-without-replacement?

btracey:
I'm not sure what you mean by 
> Is it the case that this is always wanting a sample-without-replacement?

This is weighted sampling with replacement. In a sense, this is the more fundamental concept. Weighted sampling without replacement is "Take a sample, set that weight to zero, resample". You have to shoehorn to make it work the other way ("Take a sample, set that weight to what it originally was, resample"). 

The way, then, to reduce code duplication is to change sample/Weighted is to have Weighted wrap dist.Categorical.

````
type Weighted struct {
        categorical dist.Categorical
}

func NewWeighted(w []float64, src *rand.Rand) Weighted {
     c := dist.NewCategorical(w, src)
     return Weighted{c}
}

func (w Weighted) Take() idx {
     v := w.c.Rand()
     // check if last item
     w.c.Reweight(int(v), 0)
     return int(v)
}
````
(plus others)

The main problem with this is the numeric tricks you had to play in Weighted that one does not have to play in Categorical. This seems solvable though, for example by adding methods to Categorical to support Weighted.

**Comment by kortschak** [comment]
Sorry, I missed that you don't update the weights. This algorithm is overkill if replacement is done. I'll post a comment with an alternative later today.

**Comment by kortschak** [comment]
Just looked again. There is an O(n) algorithm for non-replacement that is a fair bit simpler. This algorithm is O(logn) for each call to Rand (O(n) for New with larger constants than the O(n) above). The handling of special cases for numerical issues in Weighted is, I think, enough to justify having these as separate code.

So the question is whether the the more complex algorithm here is worthwhile (probably for high n).

LGTM

File: dist/categorical.go (hunk 3 above)
Comment (vladimir-ch) [inline dist/categorical.go]: With the comma it is easier to parse for me but still :-) Anyway, the fault is on my end, so it's ok.

File: dist/categorical.go (hunk 4 above)
Comment (vladimir-ch) [inline dist/categorical.go]: Ok

File: dist/categorical.go (hunk 5 above)
Comment (vladimir-ch) [inline dist/categorical.go]: Ok

File: dist/categorical.go [hunk 6]
```diff
@@ -0,0 +1,177 @@
...
+		last = i
+		left--
+	}
+	return float64(i - 1)
+}
+
+// Reweight sets the weight of item idx to w.
+func (c Categorical) Reweight(idx int, w float64) {
+	if w < 0 {
+		panic("categorical: negative weight")
```
vladimir-ch: Document this panic?

dist/categorical.go
```diff
@@ -0,0 +1,177 @@
...
+
+// ReweightAll resets the weights of the distribution. ReweightAll panics if
+// len(w) != c.Len.
+func (c Categorical) ReweightAll(w []float64) {
+	if len(w) != c.Len() {
+		panic("categorical: length of the slices do not match")
+	}
+	for _, v := range w {
+		if v < 0 {
+			panic("categorical: negative weight")
```
vladimir-ch: Document this panic?

dist/categorical.go
```diff
@@ -0,0 +1,177 @@
...
+
+func (c Categorical) reset() {
+	copy(c.heap, c.weights)
+	for i := len(c.heap) - 1; i > 0; i-- {
+		// Sometimes 1-based counting makes sense.
+		c.heap[((i+1)>>1)-1] += c.heap[i]
+	}
+	// TODO(btracey): Renormalization for weird weights?
+	if c.heap[0] == 0 {
+		panic("categorical: all weights equal zero")
```
vladimir-ch: Document this panic?

dist/categorical.go
```diff
@@ -0,0 +1,177 @@
...
+
+package dist
+
+import (
+	"math"
+	"math/rand"
+)
+
+// Categorical is an extension of the Bernouilli distribution where x can take
+// a certain number of possible values. Categorical must be initialized with
```
vladimir-ch: If I understand it correctly, the possible values are fixed to be `{0, 1, ..., k = len(w)-1}`. Is there any reason not to say that? Especially because you talk about zero-indexedness on the line below?

vladimir-ch:
LGTM with minor comments.

dist/categorical.go
kortschak: Thanks for reminding me of this. I was wondering why the 0-basedness was mentioned. Isn't everything we do 0-based? A description like @vladimir-ch has would be more informative.

btracey: It was mentioned because the wikipedia article defines the categorical distribution starting at 1. Doing as @vladimir-ch suggests.

dist/categorical.go
btracey: Done.

dist/categorical.go
btracey: Done.

dist/categorical.go
btracey: Done.

PR #67, opened 2015-11-17
Improve error checking for ConditionNormal

btracey:


distmv/normal.go
```diff
@@ -75,9 +75,20 @@ func NewNormal(mu []float64, sigma mat64.Symmetric, src *rand.Rand) (*Normal, bo
...
 // Mathematically this is impossible, but can occur with finite precision arithmetic.
 func (n *Normal) ConditionNormal(observed []int, values []float64, src *rand.Rand) (*Normal, bool) {
+	if len(observed) == 0 {
+		panic("normal: no observed values")
+	}
 	if len(observed) != len(values) {
 		panic("normal: input slice length mismatch")
 	}
+	if len(observed) == n.Dim() {
+		panic("normal: all dimensions observed")
```
kortschak: This is not necessarily true. If I have n{0} then I have not observed all dimensions, but this will tell me I have. You can invoke the pigeon hole principle and say that all have been observed or at least one has been observed more than once. If the check is done below the de-duplication below you can use this message. Probably just weaken the text.

btracey: Wouldn't it be more informative to have the distinct error messages (rather than weaken the text)? We have to check both individually anyway, so seems like we might as well give an informative error message.

kortschak: Sure.

btracey:
PTAL

distmv/normal.go
```diff
@@ -75,11 +75,17 @@ func NewNormal(mu []float64, sigma mat64.Symmetric, src *rand.Rand) (*Normal, bo
 // ConditionNormal returns {nil, false} if there is a failure during the update.
 // Mathematically this is impossible, but can occur with finite precision arithmetic.
 func (n *Normal) ConditionNormal(observed []int, values []float64, src *rand.Rand) (*Normal, bool) {
+	if len(observed) == 0 {
+		panic("normal: no observed values")
```
kortschak: This is an English thing, so feel free to ignore. The number of zero is singular.

I suspect that in most modern English (certainly here and in American English, but also probably in British Standard English) there are now actually three numbers, singular, plural and (new) nullar which inflects in most cases as the plural. However traditionally, the text would be "no observed value".

kortschak:
LGTM with minor comment.

distmv/normal.go
btracey: The way I typically speak, zero is plural if the number is normally plural, and singular if the number is usually singular. Here, since it accepts a slice, I pluralize it.

That said, you're right and it's fixed.

kortschak: That's an interesting observation. Number differences between ASE and BSE/AuSE are fun. An example of this is that we call an organisation plural, e.g. "gonum are developing a suite of packages" on the basis that gonum are a group of people which is plural.

PR #73, opened 2016-01-19
Make sampling routines types

btracey:


btracey:
@kortschak Here is a 10% done version of what I think you were suggesting. If this is what you were thinking I'll proceed with the full implementation.

The key interface is BatchSample and not Sample because some of the algorithms cannot generate individual samples, for example LatinHypercube. I'm happy to provide the capability for getting an individual sample from those which are IID (i.e. Importance). I'm leaning against those where the samples can be generated individually but they are not IID, for example, MetropolisHastings. I'm worried about the statefulness this would cause, and would rather that users be forced to re-initialize types.

btracey:
Could also be RandBatch / BatchRand. I'm not sure that gives the right impression, as rand is IID while this is not. It's still random though, so it's not the wrong impression.

sample/sample.go
```diff
@@ -2,8 +2,12 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// package sample contains a set of advanced routines for sampling from
+// package sample implements advanced sampling routines from explicit and implicit
```
kortschak: s/package/Package/

sample/sample.go
```diff
@@ -2,8 +2,12 @@
...
 // probability distributions.
+//
+// The sampling routines are implemented as types that satisfy the BatchSampler
+// interface. Each type has a "Func" method where the behavior is implemented
+// as a (stateless) function.
```
kortschak: Remove the parens.

kortschak:
I'm not sure of the naming. What I was thinking was something like:

```
// Sampler generates a batch of samples according to the rule specified by the
// implementing type.
type Sampler interface {
	Sample(batch []float64)
}

// LatinHypercuber is ...
type LatinHypercuber struct {
	Q   dist.Quantiler
	Src *rand.Rand
}

// Sample fills the batch with ...
func (l LatinHypercuber) Sample(batch []float64) {
	return LatinHypercube(batch, l.Q, l.Src)
}

// LatinHypercube generates len(samples) samples using Latin hypercube sampling
// from the given distribution. If src != nil, it will be used to generate
// random numbers, otherwise rand.Float64 will be used.
//
// Latin hypercube sampling divides the cumulative distribution function into equally
// spaced bins and guarantees that one sample is generated per bin. Within each bin,
// the location is randomly sampled. The dist.UnitNormal variable can be used
// for easy generation from the unit interval.
func LatinHypercube(samples []float64, q dist.Quantiler, src *rand.Rand) {
	n := len(samples)
	var perm []int
	var f64 func() float64
	if src != nil {
		f64 = src.Float64
		perm = src.Perm(n)
	} else {
		f64 = rand.Float64
		perm = rand.Perm(n)
	}
	for i := range samples {
		v := f64()/float64(n) + float64(i)/float64(n)
		samples[perm[i]] = q.Quantile(v)
	}
}
```
I don't think it's necessary to have batch in the method name unless you think that we will be having another interface that does do strict sampling. It think it's enough to have it in the docs and the signature.

btracey:
Ohh, that's what you meant. Yes, that's nicer.

PR #76, opened 2016-01-20
Add sampler interface and wrapper types for sample generation methods

btracey:


btracey:
PTAL @kortschak 

This establishes "Sample" as generate a batch of samples. I think Rand is in general not a great idea as a wrapper function, because the different algorithms have different properties for generating a single sample. Most of them cannot be used interchangeably, and thus we shouldn't provide a convenient wrapper.

The MH wrapper provides burnin and a sampling rate, while the function was not. The function does not need it because it is easily wrapped (as the implementing type shows). I think the wrapper should have it. If one wants to replace IID samples with MH samples, a burn-in time and a rate must be used. I can't see myself ever using the wrapper without wrapping the function myself to include a burn-in time. Given that, we may as well provide it. 

I think my decision is correct for Rejection. It should be an error based on how gonum defines panic behavior (that is, the behavior must be a coding error, and must be predictable beforehand). However, it is generally a really bad idea to handle this error by, for example, trying again with a higher constant. This is a statistically invalid action. Since Rejection can be used properly while wrapped we should do it, but I don't think we should handle the error, and I don't think it should panic.

sample/sample.go
```diff
@@ -42,6 +108,19 @@ func LatinHypercube(samples []float64, q dist.Quantiler, src *rand.Rand) {
 	}
 }
 
+// Importancer is a wrapper around the Importance sampling generation method.
+type Importancer struct {
+	Target   dist.LogProber
+	Proposal dist.RandLogProber
+	Src      *rand.Rand
```
sbinet: (no expertise whatsoever. take it with a grain of salt)

is `Importancer.Src *rand.Rand` really needed?
it does not seem to be used by the `Importance(...)` func (and the values are drawn from the `dist.RandLogProber`)

btracey: No, it's not, thanks. Done.

sample/sample.go
```diff
@@ -17,6 +21,68 @@ var (
...
+// WeightedSampler generates a batch of samples and their relative weights
+// according to the rule specified by the implementing type. The number of samples
+// generated is equal to the length of the input slice, and the samples and weights
+// are stored in-place into the inputs. The length of weights must equal the
+// length of samples, otherwise the WeightedSampler will panic.
+type WeightedSampler interface {
+	SampleWeighted(samples, weights []float64)
+}
+
+// WeightedSample wraps a sampler type to create a WeightedSampler where all
```
kortschak: This name is wildly misleading. Enough to confuse me about what you were doing (admittedly early in the morning). I can't think of a better name though: `UnweightedSampler`, `UniformWeightedSampler` ...

s/sampler/Sampler/

sample/sample.go
```diff
@@ -17,6 +21,68 @@ var (
...
+// weights are equal.
+type WeightedSample struct {
+	Sampler
+}
+
+func (w WeightedSample) SampleWeighted(samples, weights []float64) {
```
kortschak: Note that weights is altered by the call and how.

sample/sample.go
```diff
@@ -17,6 +21,68 @@ var (
...
+		return a
+	}
+	return b
+}
+
+// Sampler generates a batch of samples according to the rule specified by the
+// implementing type. The number of samples generated is equal to the length of
+// the input slice, and the samples are stored in-place into the input.
+type Sampler interface {
+	Sample(samples []float64)
```
kortschak: I think the parameter name `batch` works better here.

sample/sample.go
```diff
@@ -17,6 +21,68 @@ var (
...
+}
+
+// WeightedSample wraps a sampler type to create a WeightedSampler where all
+// weights are equal.
+type WeightedSample struct {
+	Sampler
+}
+
+func (w WeightedSample) SampleWeighted(samples, weights []float64) {
+	if len(samples) != len(weights) {
```
kortschak: What if `weights == nil`?

btracey: I think it can't be in this case. The point of weights is that some sampling methods MUST have a weight in order to be statistically valid. Importance sampling, for example, must have each sample corrected by its relative probability. The assumption is that if weights is not stored, the method cannot be being used properly

sample/sample.go
```diff
@@ -64,6 +145,24 @@ func Importance(samples, weights []float64, target dist.LogProber, proposal dist
 	}
 }
 
+// Rejectioner is a wrapper around the Rejection sampling generation procedure.
+// If the rejection sampling fails during the call to Sample, all samples will
+// be set to math.NaN().
+type Rejectioner struct {
```
kortschak: Rejectioner could hold an `Err error`field that can be queried after the call to `Sample`.

sample/sample.go
```diff
@@ -130,6 +229,79 @@ type MHProposal interface {
 	ConditionalRand(y float64) (x float64)
 }
 
+// MetropolisHastingser is a wrapper around the MetropolisHastings sampling type.
```
kortschak: s/sampling type/sampling function/

sample/sample.go
btracey: I thought about that, but it wasn't clear to me how it would be used. As of now, if my code takes in a Sampler, I wouldn't be able to check the Err field. If I know I'm doing rejection sample, wouldn't I call Rejection directly?

sample/sample.go
kortschak: OK. What about the comment in Importance?

sample/sample.go
kortschak: I'm not suggesting removing the current behaviour, just adding the error field. Maybe with a method so it can be generically type asserted by the caller.

sample/sample.go
btracey: Oh, oops, I'm sorry. I put it in because I asked myself the question you did, and then later decided it was a bad idea. Will remove.

sample/sample.go
btracey: Done

sample/sample.go
btracey: Made it SampleUniformWeighted.

sample/sample.go
btracey: Done.

sample/sample.go
btracey: Comment removed.

btracey:
PTAL

kortschak:
LGTM

PR #78, opened 2016-01-29
Simplified some functions

darrenmcc:
Modified some functions that I saw that weren't idiomatic Go during a cursory glance. Code is slightly simpler now. I didn't go through every file so there may be more instances that I didn't see.

### DIFF
--- a/dist/categorical.go
+++ b/dist/categorical.go
@@ -64,13 +64,12 @@ func (c Categorical) CDF(x float64) float64 {
 func (c Categorical) Entropy() float64 {
 	var ent float64
 	for _, w := range c.weights {
-		if w == 0 {
-			continue
+		if w != 0 {
+			p := w / c.heap[0]
+			ent -= p * math.Log(p)
 		}
-		p := w / c.heap[0]
-		ent += p * math.Log(p)
 	}
-	return -ent
+	return ent
 }
 
 // Len returns the number of values x could possibly take (the length of the
@@ -91,10 +90,7 @@ func (c Categorical) Mean() float64 {
 // Prob computes the value of the probability density function at x.
 func (c Categorical) Prob(x float64) float64 {
 	xi := int(x)
-	if float64(xi) != x {
-		return 0
-	}
-	if xi < 0 || xi > len(c.weights)-1 {
+	if float64(xi) != x || xi < 0 || xi > len(c.weights)-1 {
 		return 0
 	}
 	return c.weights[xi] / c.heap[0]
--- a/stat.go
+++ b/stat.go
@@ -929,18 +929,19 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
 			compensation += d
 		}
 		variance = (ss - compensation*compensation/float64(len(x))) / float64(len(x)-1)
-		return
+	} else {
+		var sumWeights float64
+		for i, v := range x {
+			w := weights[i]
+			d := v - mean
+			wd := w * d
+			ss += wd * d
+			compensation += wd
+			sumWeights += w
+		}
+		variance = (ss - compensation*compensation/sumWeights) / (sumWeights - 1)
+
 	}
 
-	var sumWeights float64
-	for i, v := range x {
-		w := weights[i]
-		d := v - mean
-		wd := w * d
-		ss += wd * d
-		compensation += wd
-		sumWeights += w
-	}
-	variance = (ss - compensation*compensation/sumWeights) / (sumWeights - 1)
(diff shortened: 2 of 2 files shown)


stat.go
```diff
@@ -76,32 +76,28 @@ func CDF(q float64, c CumulantKind, x, weights []float64) float64 {
 		return 1
 	}
 
-	var sumWeights float64
-	if weights == nil {
-		sumWeights = float64(len(x))
-	} else {
-		sumWeights = floats.Sum(weights)
-	}
+	sumWeights := floats.Sum(weights)
```
btracey: This substitution is not correct. This gives sumWeights of zero rather than len(x).

kortschak:
I'm not convinced that the majority of these changes are a win.

stat.CDF is broken by this change (probably because of the change [here](https://github.com/gonum/stat/pull/78/files#diff-33838b2a8ea2f8e4311d86efea5944c5L79) which is not correct).

stat.go
```diff
@@ -76,32 +76,28 @@ func CDF(q float64, c CumulantKind, x, weights []float64) float64 {
...
-				w += weights[i]
-			}
-		}
-		panic("impossible")
-	default:
+
+	if c != Empirical {
 		panic("stat: bad cumulant kind")
 	}
+
```
btracey: This code substitution is not desirable. Eventually there will be more cumulant kinds which will go into the switch.

stat.go
darrenmcc: tests passed, so I assumed correctness.

stat.go
```diff
@@ -934,18 +929,19 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
...
-		d := v - mean
-		wd := w * d
-		ss += wd * d
-		compensation += wd
-		sumWeights += w
 	}
-	variance = (ss - compensation*compensation/sumWeights) / (sumWeights - 1)
-	return
+
+	return mean, variance
```
kortschak: I agree with this one.

stat.go
```diff
@@ -934,18 +929,19 @@ func MeanVariance(x, weights []float64) (mean, variance float64) {
 			compensation += d
 		}
 		variance = (ss - compensation*compensation/float64(len(x))) / float64(len(x)-1)
-		return
```
btracey: I'm pretty sure returning early and reducing nesting is considered idiomatic.

dist/categorical.go
```diff
@@ -91,10 +90,7 @@ func (c Categorical) Mean() float64 {
 // Prob computes the value of the probability density function at x.
 func (c Categorical) Prob(x float64) float64 {
 	xi := int(x)
-	if float64(xi) != x {
```
kortschak: While this is logically identical, the reader-facing semantics are different for the two conditions and so are separate.

dist/categorical.go
```diff
@@ -64,13 +64,12 @@ func (c Categorical) CDF(x float64) float64 {
 func (c Categorical) Entropy() float64 {
 	var ent float64
 	for _, w := range c.weights {
-		if w == 0 {
-			continue
```
kortschak: We prefer loop continues.

stat.go
btracey: But they don't pass?

stat.go
kortschak: Yes.

darrenmcc: continue != returning early. You're really testing for (a != 0 || b != 0) to do the operation. Testing for the opposite and skipping the operation is not desirable. There is also no reduction in nesting since the if statement has to be there.

dist/categorical.go
darrenmcc: Feel free to cherry pick or disregard.

stat.go
btracey: Is your comment here supposed to refer to the change in ChiSquare instead?

kortschak: The syntactic pattern `if cond { // code; return }; // other code` is standard in Go.

stat.go
```diff
@@ -116,10 +112,9 @@ func ChiSquare(obs, exp []float64) float64 {
 	var result float64
 	for i, a := range obs {
 		b := exp[i]
-		if a == 0 && b == 0 {
```
vladimir-ch: Unrelated to this PR: is this condition correct? When `a != 0` and `b ==0`, then we get division by zero.

kortschak: Yes, this is correct.

kortschak:
Closing this.

PR #80, opened 2016-02-05
Add new Combin package for dealing with combinatorics

btracey:


btracey:
PTAL @kortschak @jonlawlor

combin/combin.go
```diff
@@ -0,0 +1,105 @@
...
+		panic(badNegInput)
+	}
+	if n < k {
+		panic(badSetSize)
+	}
+	// (n,k) = (n, n-k)
+	if k > n/2 {
+		k = n - k
+	}
+	kfact := 1
```
jonlawlor: R language uses floats for [n choose k](https://github.com/wch/r-source/blob/b156e3a711967f58131e23c1b1dc1ea90e2f0c43/src/nmath/choose.c#L110), which then allows:
```
res = n;
for i := 2; i <= k; i++ {
	    res *= (n-i+1)/i;
}
```
For k up to 29.  Not sure what source they are using for this code though.

sbinet:
naive question: isn't it a bit presomptuous to only provide an API that returns the whole slice of combinatorics?
presumably, there might be a lot of those...

I have been using this package when permutations and combinatorics have reared their ugly head:
 https://godoc.org/github.com/cznic/mathutil#PermutationFirst
 https://godoc.org/github.com/cznic/mathutil#PermutationNext

e.g.:
https://github.com/cznic/mathutil/blob/master/all_test.go#L495

jonlawlor:
@sbinet I was thinking the same thing and I'm coding up an alternative...

jonlawlor:
How about something like this:

```go

type Combos struct {
	Combo      []int
	n, k       int
	i, combins int
}

func (c Combination) Next() bool {
	if c.i >= c.combins {
		return false
	}

	for j := c.k - 1; j >= 0; j-- {
		if c.Combo[j] == c.n+j-c.k {
			continue
		}
		c.Combo[j]++
		for l := j + 1; l < c.k; l++ {
			c.Combo[l] = c.Combo[j] + l - j
		}
		break
	}
	i++
	return true
}

func GenerateCombinations(n, k int) Combos {

	data[0] = make([]int, k)
	for i := range data {
		data[i] = i
	}
	return Combos{
		Combo:   data,
		n:       n,
		k:       k,
		combins: Binomial(n, k)}

}```

File: combin/combin.go (hunk 1 above)
Comment (btracey) [inline combin/combin.go]: That is definitely better on the upper end for not encountering overflow, but what about rounding error kinds of things? I'm inclined to keep the exact calculation exact, and users can use GeneralizedBinomial if they want inexact with rounding.

**Comment by btracey** [comment]
It seems to me that both are useful. The explicit construction is easier to parallelize and do things like cross-validation. Obviously the generator is better for memory. Just need to find naming in my opinion.

**Comment by jonlawlor** [comment]
The explicit construction could call the generator by adding a call to copy in a loop.  It would use 1 more slice of memory.

**Comment by btracey** [comment]
Implemented a variant of your idea. There is now an explicit constructor and a generator as suggested.

PTAL

**Comment by jonlawlor** [comment]
I like it, that is cleaner and clearer than I was suggesting. LGTM

**Comment by btracey** [comment]
Just that section, or is that a full code review lgtm?

**Comment by jonlawlor** [comment]
Full code.  I assume this is where things like permutations would also go here?  The reason I ask is because names could be shorter if not.

File: combin/combin.go [hunk 2]
```diff
@@ -0,0 +1,169 @@
...
+		num *= i
+	}
+	return num / kfact
+}
+
+// GeneralizedBinomial returns the generalized binomial coefficient of (n, k).
+//  Γ(n+1) / (Γ(k+1) Γ(n-k+1))
+// where Γ is the gamma function. GeneralizedBinomial is useful for continuous
+// relaxations of the binomial coefficient, or when the binomial coefficient value
+// may overflow int. In the latter case, one may alternately use math/big for
```
kortschak: alternatively

btracey:
Yea, I was thinking permutations would go here too. It didn't feel like it belonged anywhere else, so I created a new package.

combin/combin.go
```diff
@@ -0,0 +1,169 @@
...
+func NewCombinationGenerator(n, k int) *CombinationGenerator {
+	combins := Binomial(n, k)
+	return &CombinationGenerator{
+		n:         n,
+		k:         k,
+		remaining: combins,
+	}
+}
+
+// Remaining returns the number of combinations remaining to be generated.
```
kortschak: Is there a reason a caller would need to know how many remain as opposed to knowing the call to Next will not fail.

I would rename this to `Next() bool` and rename `Next` to `Combination`.

combin/combin.go
```diff
@@ -0,0 +1,169 @@
...
+func (c *CombinationGenerator) Remaining() int {
+	return c.remaining
+}
+
+// Next generates the next combination. If next is non-nil, it must have length
```
kortschak: This documentation should say how next is indexed into. Possibly there should be a cursor method.

combin/combin.go
```diff
@@ -0,0 +1,169 @@
...
+	badSetSize  = "combin: n < k"
+	badInput    = "combin: wrong input slice length"
+)
+
+// Binomial returns the binomial coefficient of (n,k).
+//
+// The binomial coefficient is the number of unordered combinations of k elements
+// in a set that is n elements big. It is also commonly referred to as "n choose k".
+//
+//  n!/((n-k)!k!)
```
vladimir-ch: I like the formulas to read as a part of the text. How about restructuring the comment as:
```
// Binomial returns the binomial coefficient of (n,k), also commonly referred to as "n choose k".
//
// The binomial coefficient C(n,k) is the number of unordered combinations of k elements
// in a set that is n elements big, and is defined as
//
//  C(n,k) := n!/((n-k)!k!).
//
// n and k must be non-negative with n >= k, otherwise Binomial will panic.
// No check is made for overflow.
```
Just a suggestion.

combin/combin.go
```diff
@@ -0,0 +1,169 @@
...
+		kfact *= i
+	}
+	num := 1
+	for i := n; i > n-k; i-- {
+		num *= i
+	}
+	return num / kfact
+}
+
+// GeneralizedBinomial returns the generalized binomial coefficient of (n, k).
```
vladimir-ch: Also here, how about
```
// GeneralizedBinomial returns the generalized binomial coefficient of (n, k)
// defined as
//  Γ(n+1) / (Γ(k+1) Γ(n-k+1)),
// where Γ is the gamma function. ...
```
?

combin/combin.go
```diff
@@ -0,0 +1,169 @@
...
+// CombinationGenerator generates combinations iteratively. Combinations may be
+// called to generate all combinations collectively.
+type CombinationGenerator struct {
+	n         int
+	k         int
+	previous  []int
+	remaining int
+}
+
+// NewCombinationGenerator returns a CombinationGenerator for a generating the
```
vladimir-ch: ... for ~~a~~ generating the ...

combin/combin.go
```diff
@@ -0,0 +1,169 @@
...
+	return next
+}
+
+// Combinations generates all of the combinations of k elements from a
+// set of size n. The returned slice has length Binomial(n,k) and each inner slice
+// has length k.
+//
+// n and k must be non-negative with n >= k, otherwise Combinations will panic.
+//
+// CombinationGenerator may alternately used to generate each combination
```
vladimir-ch: `s/alternately/alternatively/`.
Also, `be` is missing.
And how about `s/each combination/the combinations/` ?

combin/combin_test.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+		}
+	}
+	return true
+}
+
+var binomialTests = []struct {
+	n, k, ans int
+}{
+	{5, 0, 1},
```
vladimir-ch: Add `{0, 0, 1}` ?

combin/combin.go
btracey: Done.

combin/combin.go
btracey: Done.

combin/combin.go
btracey: Done.

combin/combin.go
btracey: Done.

combin/combin.go
btracey: I was thinking one may want to know the total number of combinations (if a function is passed a Generator without being given n and k). This covers both cases.

btracey: Can change though if you'd like.

combin/combin.go
btracey: Done.

combin/combin_test.go
btracey: Done.

btracey:
Changes made, PTAL

combin/combin.go
kortschak: See what others think. I'm happyish with the remaining, but I don't like that you can say `for c.Remaining() > 0 {}` and not exit the loop. I'd be much happier if the loop clause contained the advance.

btracey: Okay, that's fair.

vladimir-ch: I'm also slightly leaning towards @kortschak 's suggestion.

btracey:
Changed to "Next() bool" in CombinationGenerator

combin/combin.go
```diff
@@ -0,0 +1,171 @@
...
+// combinations of k elements from a set of size n.
+//
+// n and k must be non-negative with n >= k, otherwise NewCombinationGenerator
+// will panic.
+func NewCombinationGenerator(n, k int) *CombinationGenerator {
+	combins := Binomial(n, k)
+	return &CombinationGenerator{
+		n:         n,
+		k:         k,
+		remaining: combins,
```
vladimir-ch: Call `Binomial` directly?

btracey: Done.

vladimir-ch:
LGTM

kortschak:
I'd still like to see the advancement of the iterator put into the Next method.

btracey:
Presumably still extracted with Combination?

kortschak:
Yes.

btracey:
Fixed. PTAL.

kortschak:
What is the expected behaviour if a caller calls Combination before calling Next? Currently it is silent.

There should probably be a note on Combination or Next (or both) that the result of a call to Combination will remain the same until Next is called again – I’m seeing this as something akin to the validity of a shared buffer between reads.

btracey:
Good point. I would say it should panic. Will add the comment you suggest.

btracey:
Done.

kortschak:
Final quibble. Add that Combination will panic otherwise, then LGTM.

combin/combin.go
```diff
@@ -0,0 +1,187 @@
...
+	previous  []int
+	remaining int
+}
+
+// NewCombinationGenerator returns a CombinationGenerator for generating the
+// combinations of k elements from a set of size n.
+//
+// n and k must be non-negative with n >= k, otherwise NewCombinationGenerator
+// will panic.
+func NewCombinationGenerator(n, k int) *CombinationGenerator {
```
sbinet: (apologies for joining late)
any reason why you return a pointer?
it's still reasonnably small (3ints+3words)...

kortschak: The methods are on the pointer since it self mutates.

sbinet: ah, right!

PR #82, opened 2016-02-07
Change random seed to go default

jonlawlor:
Go default for the random seed is 1, not 0.  We should use the default unless there is a reason not to.


btracey:
LGTM

jonlawlor:
Not LGTM, the test is dependent on the ordering that rand.Seed(0) provides.  I think I'll delete this, there is nothing wrong with using a non default random seed, as long as there is a reason.

PR #85, opened 2016-02-10
Add NewNormalPrecision for constructing a normal distribution based o…

btracey:
…n its precision matrix

### DIFF
--- a/distmv/normal.go
+++ b/distmv/normal.go
@@ -77,6 +77,34 @@ func NewNormalChol(mu []float64, chol *mat64.Cholesky, src *rand.Rand) *Normal {
 	return n
 }
 
+// NewNormalPrecision creates a new Normal distribution with the given mean and
+// precision matrix (inverse of the covariance matrix). NewNormalPrecision
+// panics if len(mu) is not equal to prec.Symmetric(). If the precision matrix
+// is not positive-definite, NewNormalPrecision returns nil for norm and ok
+// for false.
+func NewNormalPrecision(mu []float64, prec *mat64.SymDense, src *rand.Rand) (norm *Normal, ok bool) {
+	if len(mu) == 0 {
+		panic(badZeroDimension)
+	}
+	dim := prec.Symmetric()
+	if dim != len(mu) {
+		panic(badSizeMismatch)
+	}
+	// TODO(btracey): Computing a matrix inverse is generally numerically instable.
+	// This only has to compute the inverse of a positive definite matrix, which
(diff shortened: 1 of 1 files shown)


btracey:
Needs https://github.com/gonum/matrix/pull/333

distmv/normal.go
```diff
@@ -77,6 +77,33 @@ func NewNormalChol(mu []float64, chol *mat64.Cholesky, src *rand.Rand) *Normal {
 	return n
 }
 
+// NewNormalPrecision creates a new Normal distribution with the given mean and
+// precision matrix (inverse of the covariance matrix). NewNormalPrecision
+// panics if len(mu) is not equal to prec.Symmetric(). If the precision matrix
+// is not positive-definite, NewNormalPrecision returns {nil,false}.
+func NewNormalPrecision(mu []float64, prec *mat64.SymDense, src *rand.Rand) (*Normal, bool) {
```
kortschak: Name the returns.

distmv/normal.go
```diff
@@ -77,6 +77,33 @@ func NewNormalChol(mu []float64, chol *mat64.Cholesky, src *rand.Rand) *Normal {
 	return n
 }
 
+// NewNormalPrecision creates a new Normal distribution with the given mean and
+// precision matrix (inverse of the covariance matrix). NewNormalPrecision
+// panics if len(mu) is not equal to prec.Symmetric(). If the precision matrix
+// is not positive-definite, NewNormalPrecision returns {nil,false}.
```
kortschak: Rephrase the return to take into account return parameter names and avoid braces.

kortschak:
Is there a reason to do all this rather than let the client find the inverse (perhaps via a solve)? I'm happy if you think so, then LGTM.

btracey:
There are two reasons:
1) It seems better to have the user say what they want and we do it the right way then to ask for more.
2) It allows more flexibility for changing. If we find this is the wrong way to do it (http://mathoverflow.net/questions/230808/computing-the-inverse-of-a-cholesky-decomposition), clients get the upgraded algorithm for free. Similarly, if we decide storing both is worth it (it may be worth it for machine-learning like applications), clients get that for free too. 

If that seems a good enough reason I'll merge when changes are implemented.

kortschak:
Those arguments are fine. Merge away!

distmv/normal.go

distmv/normal.go

kortschak:
s/ok for false/false for ok/

PR #88, opened 2016-02-29
stat: add principal components calculation

kortschak:
@btracey @vladimir-ch @Armadilloa16 Please take a look.


Armadilloa16:
Because of the centring step,
lines 25-27 of pca.go should be 
if k > (d-1) {
       panic("stat: k > columns-1")
}

Otherwise LGTM.

Got some suggestions for possible improvements
though:

I would suggest you add tests for non-square 
matrices, in both ways (n > d and n < d) and in the 
n > d case include one for which the covariance 
matrix is non-singular. 

I would also consider not including (or calculating?) 
PC's corresponding to singular values of zero -- 
although calculating this might be tricky. 
One possible simplification/ improvement could be 
to first calculate the rank r of the input a, then you 
need only check that k <= r instead of the two 
checks, and/or you could only include/calculate 
the first r PCs and singular values if k > r.

kortschak:
>I would suggest you add tests for non-square
matrices, in both ways (n > d and n < d) and in the
n > d case include one for which the covariance
matrix is non-singular.

If you can provide some good cases I'll add them.

Armadilloa16:
> If you can provide some good cases I'll add them.

Some cases:

A portion of Fisher's (Anderson's) Iris data (I've simply taken the Iris-setosa cases
but it would actually be useful to have the entire dataset somewhere for testing
classification stuff whenever I or someone else gets around to adding that).
The Iris data is a canonical dataset in classification. This is a typical n > d 
non-singular example.
[5.1,3.5,1.4,0.2;
4.9,3.0,1.4,0.2;
4.7,3.2,1.3,0.2;
4.6,3.1,1.5,0.2;
5.0,3.6,1.4,0.2;
5.4,3.9,1.7,0.4;
4.6,3.4,1.4,0.3;
5.0,3.4,1.5,0.2;
4.4,2.9,1.4,0.2;
4.9,3.1,1.5,0.1;
5.4,3.7,1.5,0.2;
4.8,3.4,1.6,0.2;
4.8,3.0,1.4,0.1;
4.3,3.0,1.1,0.1;
5.8,4.0,1.2,0.2;
5.7,4.4,1.5,0.4;
5.4,3.9,1.3,0.4;
5.1,3.5,1.4,0.3;
5.7,3.8,1.7,0.3;
5.1,3.8,1.5,0.3;
5.4,3.4,1.7,0.2;
5.1,3.7,1.5,0.4;
4.6,3.6,1.0,0.2;
5.1,3.3,1.7,0.5;
4.8,3.4,1.9,0.2;
5.0,3.0,1.6,0.2;
5.0,3.4,1.6,0.4;
5.2,3.5,1.5,0.2;
5.2,3.4,1.4,0.2;
4.7,3.2,1.6,0.2;
4.8,3.1,1.6,0.2;
5.4,3.4,1.5,0.4;
5.2,4.1,1.5,0.1;
5.5,4.2,1.4,0.2;
4.9,3.1,1.5,0.1;
5.0,3.2,1.2,0.2;
5.5,3.5,1.3,0.2;
4.9,3.1,1.5,0.1;
4.4,3.0,1.3,0.2;
5.1,3.4,1.5,0.2;
5.0,3.5,1.3,0.3;
4.5,2.3,1.3,0.3;
4.4,3.2,1.3,0.2;
5.0,3.5,1.6,0.6;
5.1,3.8,1.9,0.4;
4.8,3.0,1.4,0.3;
5.1,3.8,1.6,0.2;
4.6,3.2,1.4,0.2;
5.3,3.7,1.5,0.2;
5.0,3.3,1.4,0.2]

A singular n > d case could easily be constructed by 
taking some of these rows, something like
[5.1,3.5,1.4,0.2;
4.9,3.0,1.4,0.2;
4.7,3.2,1.3,0.2;
4.6,3.1,1.5,0.2;
5.0,3.6,1.4,0.2;
5.1,3.8,1.6,0.2;
4.6,3.2,1.4,0.2;
5.3,3.7,1.5,0.2;
5.0,3.3,1.4,0.2]

You could use the matrix transpose of the above two examples
as n < d cases.

Armadilloa16:
> Because of the centring step,
> lines 25-27 of pca.go should be 
> if k > (d-1) {
> panic("stat: k > columns-1")
> }

My bad I got confused. Sorry. What I meant was:
Because of the centreing step,
lines 22-24 of pca.go should be 
if k > n - 1 {
	panic("stat: k > rows - 1")
 }

pca.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/gonum/matrix"
+	"github.com/gonum/matrix/mat64"
+)
+
+// PrincipalComponents returns the k principal component direction vectors of
+// the matrix a, projected score vectors and variance values, where variables
+// are represented in the columns and observations are the rows in the input.
+// The principal component direction vectors are found by singular value
+// decomposition of the input.
+// If no singular value decomposition is possible, pc, proj and sigma are returned
```
vladimir-ch: s/sigma/vars/

pca_test.go
```diff
@@ -0,0 +1,166 @@
...
+package stat
+
+import (
+	"testing"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix/mat64"
+)
+
+func TestPrincipleComponents(t *testing.T) {
```
vladimir-ch: s/Principle/Principal/

pca.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	var c, v mat64.Dense
+	v.VFromSVD(&svd)
+	pc = v.View(0, 0, d, k).(*mat64.Dense)
+	c.Mul(centered, pc)
+	vars = svd.Values(nil)[:k]
+	f = 1 / float64(n-1)
+	for i, v := range vars {
+		vars[i] *= v * f
```
vladimir-ch: Wouldn't this be marginally clearer as `vars[i] = f * v * v`?

vladimir-ch:
LGTM

pca.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/gonum/matrix/mat64"
+)
+
+// PrincipalComponents returns the k principal component direction vectors of
+// the matrix a, projected score vectors and variance values, where variables
+// are represented in the columns and observations are the rows in the input.
+// The principal component direction vectors are found by singular value
+// decomposition of the input.
+// If no singular value decomposition is possible, pc, proj and sigma are returned
+// nil and ok is returned false. k must not be greater than the number of rows less
```
jonlawlor: I would rephrase as `PrincipalComponents will panic if k is greater than or equal to the number of rows, or if k is greater than the number of columns.`

btracey:
Is this the right place for it? Active subspaces (http://activesubspaces.org/research/) is a very similar technique. Would it also go into stat or do we want a "dimension reduction" kind of package? I think it could be nice to have routines for a) performing the decomposition, b) selecting the number of dimensions (at the start just user supplied) c) projection functions into and out of transformed space.

pca.go
```diff
@@ -0,0 +1,63 @@
...
+
+package stat
+
+import (
+	"github.com/gonum/matrix"
+	"github.com/gonum/matrix/mat64"
+)
+
+// PrincipalComponents returns the k principal component direction vectors of
+// the matrix a, projected score vectors and variance values, where variables
```
btracey: This sentence could be broken. First is what it does, the second describes the input.

pca.go
```diff
@@ -0,0 +1,63 @@
...
+// are represented in the columns and observations are the rows in the input.
+// The principal component direction vectors are found by singular value
```
btracey: Do we want to say this? I imagine there could exist a fast technique for finding the first k vectors (or what have you).

pca.go
```diff
@@ -0,0 +1,63 @@
...
+// decomposition of the input.
+// If no singular value decomposition is possible, pc, proj and sigma are returned
+// nil and ok is returned false. k must not be greater than the number of rows less
+// one or the number of columns in a, or PrincipalComponents will panic.
+func PrincipalComponents(a mat64.Matrix, k int) (pc, proj *mat64.Dense, vars []float64, ok bool) {
```
btracey: I wonder if there's an easy way to make this weighted.

jonlawlor: I imagine you could do the same as in covariancematrix, and multiply each row by the square root of the weight?

jonlawlor: Turns out my gut was right: http://stats.stackexchange.com/questions/113485/weighted-principal-components-analysis

btracey:
To clarify the last comment, I don't think we need to design a whole dimension reduction package at the moment, just noting that we may want to move or modify it later.

pca.go
kortschak: Yes, habit. Here the dominant impact is not this loop. Done.

pca_test.go
kortschak: Done.

pca.go
kortschak: Done.

pca.go
kortschak: Done.

pca.go
kortschak: Done.

pca.go
kortschak: It is to distinguish from the alternative ED approach which is less stable.

pca.go
kortschak: Done.

kortschak:
I've addressed the majority of comments. There are some tests that I want to add, but I want to think about this more.

The reasoning behind putting it here is that CovarianceMatrix is here are they are related.

PTAL

pca.go
kortschak: On further consideration, I don't think this is right. We should allow scaling of the variables, not the observations.

btracey: I can’t see on github where this comment exists. Is it in reference to the new weighted API?

kortschak: You need to go via the show outdated commit link on the conversation page. The specific comment starts from your third line comment about weighting.

The github UI for this is awful.

btracey: Aha!

I disagree. Weighted values can come up for a variety of valid reasons. There could be actual duplicates in a long table, and allowing weights can reduce the size of the input matrix. Another reason could be input locations are generated by a method like Importance Sampling (or more advanced techniques like cross-entropy) and the data points must be given different relative importances. Finally, they could come up in some sort of "goal oriented" PCA where one is interested in local importance directions (like in locally weighted least squares) and where the weighting function is defined by some sort of kernel and not just a threshold function.

kortschak: OK. Then we should at least allow standardisation as well. Currently we don't. `scale []float64` for the variables?

btracey: You're talking about the variable sizes? That is, having something like all the columns have a variance of one? I agree that's a useful thing to have, but it feels orthogonal to the PCA call. The variable scales can be changed independently by multiplying individual variable elements by the correct number.

Note that the same is not true for data weights. Data weighting cannot be accomplished by just multiplying specific values by a number. In PCA and covariance matrix, it has to come exactly after scaling to mean zero and has to use the square root of the weights, while in other algorithms the weights come in linearly (say, mean).

kortschak: Yes, that's what I mean. Scaling is not by a specific number though, it's also by a vector. R does this for example.

kortschak: I'll push what I'm suggesting under the understanding that I'll revert it if people are not happy.

btracey: I understand that (I think). The point remains that rescaling data is a pre-processing step while considering weighted data is not. This means that weights should be a part of the PCA signature while scaling should not. 

One can do (a la https://godoc.org/github.com/reggo/reggo/scale) 
````
scaledData := scaler.ScaleData(nil, data)
stat.PCA(scaledData, nil)
````
Scaling should thus be kept as orthogonal to the PCA algorithm.

Comment (kortschak) [inline pca.go]: OK

File: pca.go [hunk 8]
```diff
@@ -0,0 +1,76 @@
...
+	"math"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix"
+	"github.com/gonum/matrix/mat64"
+)
+
+// PrincipalComponents returns the k principal component direction vectors of
+// the matrix a projected score vectors and variance values by singular value
+// decomposition of the input. The input a is a matrix where variables are
```
btracey: Bikeshedding, but "The input a is a matrix where each observation is a row and each column represents a variable"?

pca.go
```diff
@@ -0,0 +1,76 @@
...
+// represented in the columns and observations are the rows in the input.
+//
+// If weights is not nil, it is used to weight the observations. The length of
```
btracey: Weights is used to weight the observations. If weights is nil, each weight is considered to have a value of one, otherwise the length of weight must match ...

pca.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"math"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix"
+	"github.com/gonum/matrix/mat64"
+)
+
+// PrincipalComponents returns the k principal component direction vectors of
+// the matrix a projected score vectors and variance values by singular value
```
btracey: Comma after a? If not that, then some other way of making that first sentence more clear.

pca.go
kortschak: Much nicer.

btracey: It would be nice to have a mathematical (or at least more in-depth) description in between the current first sentence and the description of the inputs. I even know what PCA does, and it's not clear from the documentation what "pc", "proj", and "vars" are. Such a description should at least say what the sizes of the outputs are.

pca.go
kortschak: Done.

pca_test.go
```diff
@@ -0,0 +1,282 @@
...
+func TestPrincipalComponents(t *testing.T) {
+	for i, test := range []struct {
+		data     mat64.Matrix
+		k        int
+		wantPC   *mat64.Dense
+		wantProj *mat64.Dense
+		wantVars []float64
+		epsilon  float64
+	}{
+		// Test results manually verified using octave.
```
btracey: Is it hard to add a manually verified result with a full-rank matrix?

kortschak: If you could provide a case (also a weighted test case would be helpful), I'll add it.

pca_test.go
```diff
@@ -0,0 +1,282 @@
...
+				-0.1300569, -0.009838325, -0.074538231,
+				-0.1519997, -0.047089299, -0.022845757,
+				-0.1414906, 0.048150093, 0.043540423,
+			}),
+			wantProj: nil,
+			wantVars: []float64{222.78428394075226, 0.8252354865184671, 0.2821472393958653},
+			epsilon:  1e-6, // Reduced precision because PC vectors are obtained from R by copy/paste.
+		},
+	} {
+		pc, proj, vars, ok := PrincipalComponents(test.data, nil, test.k)
```
btracey: We should have a test with weighted data. I can add that later if you'd prefer given that the addition is my fault :-).

In either event, it's better for weights to be an element in test rather than a nil literal.

kortschak: Done while you were writing that.

pca.go
```diff
@@ -0,0 +1,76 @@
...
+	if k > d {
+		panic("stat: k > variables")
+	}
+	if weights != nil && len(weights) != n {
+		panic("stat: len(weights) != observations")
+	}
+
+	centered := mat64.NewDense(n, d, nil)
+	f := 1 / float64(n)
+	for j := 0; j < d; j++ {
```
btracey: This is wrong with weighted data. The mean also needs to be weighted. I think this is fixed with 

````
col := make([]float64,n)
for j := 0; j < d; j++ {
     mat64.Col(col, a, j)
     mean := Mean(col, weights)
     floats.AddConst(col, -mean)
     centered.SetCol(j, col)
}

Comment (kortschak) [inline pca.go]: I was going from the link @jonlawlor provided, but I'll defer to you.

Comment (btracey) [inline pca.go]: Which says "First, one needs to compute the weighted mean μ=1n∑wixi and subtract it from the data in order to center it."

File: pca.go [hunk 14]
```diff
@@ -0,0 +1,77 @@
...
+	if !ok {
+		return nil, nil, nil, false
+	}
+
+	var c, v mat64.Dense
+	v.VFromSVD(&svd)
+	pc = v.View(0, 0, d, k).(*mat64.Dense)
+	c.Mul(centered, pc)
+	vars = svd.Values(nil)[:k]
+	f = 1 / float64(n-1)
```
btracey: This needs to have ```f = 1 /  (floats.Sum(weights) - 1)``` if weights != nil

pca_test.go
btracey: Will do. I have to go down the software rabbit hole apparently, but working on it.

pca.go
```diff
@@ -0,0 +1,77 @@
...
+//
+// The slice weights is used to weight the observations. If weights is nil,
+// each weight is considered to have a value of one, otherwise the length of
+// weight must match the number of observations or PrincipalComponents will panic.
+//
+// If no singular value decomposition is possible, pc, proj and vars are returned
+// nil and ok is returned false. PrincipalComponents will panic if k is greater
+// than or equal to the number of rows, or if k is greater than the number of
+// columns.
+func PrincipalComponents(a mat64.Matrix, weights []float64, k int) (pc, proj *mat64.Dense, vars []float64, ok bool) {
```
btracey: Does "proj" actually need to be an output? The main outputs of PCA are the singular values and vectors. The final projection step is just a niceness. I'm not sure I'm remembering my PCA correctly, but usually one is also interested in the projections of new points, which one computes with  Y_new = pc * X_new. However, the code here computes that projection on the zero mean. Thus, new points won't have the same zero as the returned projection. It seems much better to just return pc and vars, and then make a note in the documentation about how projections can be done.

pca.go
kortschak: Ah yes, sorry.

kortschak: Done (calls needed modification of parameter order).

pca.go
kortschak: Done.

pca.go
kortschak: The reason I included it is that the projection is done on the centred and weighted data, which we already have. Otherwise the client needs to construct that again to get the projection.

kortschak: I see your point though. @Armadilloa16 what do you think?

pca.go
```diff
@@ -0,0 +1,81 @@
...
+	"math"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix"
+	"github.com/gonum/matrix/mat64"
+)
+
+// PrincipalComponents returns the k principal component direction vectors of
+// the matrix a, projected score vectors and variance values by singular value
+// decomposition of the input. The input a is a matrix where each observation
```
vladimir-ch: Make it `... where each row is an observation ... `? Would be consistent with `... each column represents a variable.`.

kortschak: Done.

pca.go
kortschak: Taking this approach further, I can drop k as an input since it doesn't actually save any space (the view still has the backing matrix), and return all the PC vectors and all the variances.

btracey: That's fine with me. There's usually a second processing step where the number of components is selected (unless it's predecided to be 1 or 2 or something)

kortschak:
*ping*

btracey:
Code looks very nice. Checking some things on documentation.

pca.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"math"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix"
+	"github.com/gonum/matrix/mat64"
+)
+
+// PrincipalComponents returns the principal component direction vectors of
```
btracey: The documentation is in general good. Two comments
1) PCA freqently says that the columns need to be zero mean and variance of 1. This variance is not rescaled. This is the correct approach, but I think it should be noted.
2) I don't see anything about the singular values changing from v --> f * v^2. I'm sure this is a standard thing, but I don't see any mention of it in the wikipedia article or in the Stanford machine learning notes. Could you provide a link that explains "variance values" so that users can read what it means?

Armadilloa16: I agree you should note the behaviour about centring and scaling to variance 1.

Also @btracey I think I can help clarify the singular value thing.
The v -> f*v^2 results from the fact we are calculating the singular values of 
the data matrix a. This is sort of briefly covered on the wiki page:
https://en.wikipedia.org/wiki/Principal_component_analysis#Singular_value_decomposition
The standard formulation of PCA instead involves the 
eigendecomposition of the covariance matrix, (1/(n-1)) * a^T * a (assuming the 
columns of a are centred). The eigenvalues of this covariance matrix will be 
directly related to the singular values of a by the above transformation. This can 
be seen reasonably easily, if the singular value decomposition of a is written:
a = U * S * V^T
then the covariance matrix becomes
(1/(n-1)) * a^T * a  = (1/(n-1)) * (U * S * V^T)^T * (U * S * V^T) =  V * ((1/(n-1)) * S^2) * V^T
I cant find a handy link to this, but it is explained in several multivariate 
stats textbooks, including this one:
http://www.cambridge.org/au/academic/subjects/statistics-probability/statistical-theory-and-methods/analysis-multivariate-and-high-dimensional-data?format=HB
Hopefully that helps?

btracey: Yep, of course, thanks.

I'm not sure how to phrase it. I think "and variance values by singular value decomposition" is misleading, because it kind of suggests the singular values themselves. Maybe "PCA returns the principal component direction vectors and variance values of the matrix a, computed using the SVD of the input"? That at least puts some space between what is being returned (the variance value) and how it's being computed.

Armadilloa16: You could say "... returns the principle component direction vectors and the variances of the principle component scores, vecs * a, computed using the SVD of the input a" as vars are infact the variances of the columns of vecs * a (where * is matrix multiplication, and assuming a is centred already).

kortschak:
Documentation changes made. Do we have any weighted test cases?

PTAL

pca.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"math"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix"
+	"github.com/gonum/matrix/mat64"
+)
+
+// PrincipalComponents returns the principal component direction vectors and
+// the variances of the principle component scores, vecs * a, computed using
```
btracey: The diagonal of vecs * a? vecs * a is a d x min(n,d) matrix, while vars is one dimensional.

btracey:
Here are some weighted tests. This code has the right numbers but the wrong sign for some of the vectors (at least according to octave). It should be okay since SVD isn't unique. I confirmed that the SVD matched by adding 

````
var u mat64.Dense
		u.UFromSVD(&svd)
		diag := mat64.NewDense(len(vars), len(vars), nil)
		for i := range vars {
			diag.Set(i, i, vars[i])
		}
		var ans mat64.Dense
		ans.Product(&u, diag, v.T())
		fmt.Println("equal approx", mat64.EqualApprox(&ans, centered, 1e-8))
```

Test structs:
````
{ // Truncated iris data unitary weights.
			data: mat64.NewDense(10, 4, []float64{
				5.1, 3.5, 1.4, 0.2,
				4.9, 3.0, 1.4, 0.2,
				4.7, 3.2, 1.3, 0.2,
				4.6, 3.1, 1.5, 0.2,
				5.0, 3.6, 1.4, 0.2,
				5.4, 3.9, 1.7, 0.4,
				4.6, 3.4, 1.4, 0.3,
				5.0, 3.4, 1.5, 0.2,
				4.4, 2.9, 1.4, 0.2,
				4.9, 3.1, 1.5, 0.1,
			}),
			weights: []float64{1, 1, 1, 1, 1, 1, 1, 1, 1, 1},
			wantVecs: mat64.NewDense(4, 4, []float64{
				-0.6681110197952722, 0.7064764857539533, -0.14026590216895132, -0.18666578956412125,
				-0.7166344774801547, -0.6427036135482664, -0.135650285905254, 0.23444848208629923,
				-0.164411275166307, 0.11898477441068218, 0.9136367900709548, 0.35224901970831746,
				-0.11415613655453069, -0.2714141920887426, 0.35664028439226514, -0.8866286823515034,
			}),
			wantVars: []float64{0.1665786313282786, 0.02065509475412993, 0.007944620317765855, 0.0019327647109368329},
			epsilon:  1e-12,
		},
		{ // Truncated iris data non-unitary weights.
			data: mat64.NewDense(10, 4, []float64{
				5.1, 3.5, 1.4, 0.2,
				4.9, 3.0, 1.4, 0.2,
				4.7, 3.2, 1.3, 0.2,
				4.6, 3.1, 1.5, 0.2,
				5.0, 3.6, 1.4, 0.2,
				5.4, 3.9, 1.7, 0.4,
				4.6, 3.4, 1.4, 0.3,
				5.0, 3.4, 1.5, 0.2,
				4.4, 2.9, 1.4, 0.2,
				4.9, 3.1, 1.5, 0.1,
			}),
			weights: []float64{2, 3, 1, 1, 1, 1, 1, 1, 1, 2},
			wantVecs: mat64.NewDense(4, 4, []float64{
				-0.618936145422414, 0.763069301531647, 0.124857741232537, 0.138035623677211,
				-0.763958271606519, -0.603881770702898, 0.118267155321333, -0.194184052457746,
				-0.143552119754944, 0.090014599564871, -0.942209377020044, -0.289018426115945,
				-0.112599271966947, -0.212012782487076, -0.287515067921680, 0.927203898682805,
			}),
			wantVars: []float64{0.129621985550623, 0.022417487771598, 0.006454461065715, 0.002495076601075},
			epsilon:  1e-12,
		},
````

btracey:
LGTM with tests added and comment fixed (or explained to me :) ) .

pca.go
kortschak: The *column* variances of vecs*a. I'll add that word and it should be clear.

btracey: Ahhh, yes. Very clear

kortschak:
PTAL

PR #89, opened 2016-03-03
Addexpected

btracey:


btracey:
I do think this function belongs in gonum, even though it's a wrapper around quad.Fixed.
a) The integral substitution is non-obvious and is significantly better than the naive algorithm
b) When quad gains other polynomial classes, this function can get significantly better. For example, the variance of the unit normal takes 1,000,000 quadrature points with Legendre to get to 1e-10, but should only take 2 with Hermite polynomials (it's an integral of a polynomial over a Gaussian). It's nice to have a place that automatically gets better as Quad does.

I'm not sure this is the correct place, but I think it is.
a) I don't think quad needs to import dist, so we don't need to worry about circular imports
b) As per b) above, we're likely to do type switches in the future. It seems most natural to do such type switches in the package where the types live.

kortschak:
Presumably you mean belongs in stat. Yes, I agree.

dist/expected.go
```diff
@@ -0,0 +1,54 @@
...
+// is, ExpectedFixed computes
+//
+//  ev ≈ E[f] = int_-inf^inf f(x) p(x) dx
+//
+// using the number of quadrature locations specified by evals. Concurrent
+// sets the number of allowed concurrent evaluations of f.
+//
+// ExpectedFixed is a wrapper around quad.Fixed. See the documentation for more
+// information.
+func ExpectedFixed(f func(float64) float64, q Quantiler, evals, concurrent int) (ev float64) {
```
vladimir-ch: Is it necessary to require `concurrent`? I know that it is for `quad.Fixed` but seeing it in `dist` seems a bit out of place, too low level and unrelated to expected value.

dist/expected.go
```diff
@@ -0,0 +1,54 @@
...
+	// p(x) ≈ 0. Additionally, the original integral cannot be computed numercally
+	// for discrete distributions, as p(x) is a series of 𝛿-functions, but can
+	// when using the quantile.
+
+	// TODO(btracey): Use a type switch on the quantiler when quad supports
+	// more quadrature classes. For example, use Gaussian-Hermite quadrature
+	// for the normal distribution.
+	// TODO(btracey): Add a test for a categorical distribution when one that
+	// is a Quantiler exists.
+	g := func(p float64) float64 {
```
vladimir-ch: Is this substitution really so non-trivial to justify this function? It looks like first-year calculus. I'm really just asking.

dist/expected.go
kortschak: I think it might be reasonable to set this to min(evals, runtime.GOMAXPROCS(0)).

btracey: The problem is that in many non-trivial cases it's much harder to write race-free functions (unless you've been planning on concurrency from the start). If we think it's easier to be a boolean (like in concurrent) that's fine with me, but I do think it should be opt-in because of race concerns.

dist/expected.go
btracey: I think it's non-obvious, even though it's trivial once you know you can do it. That may not be enough on its own, but I think future type switching makes it worth it. Alternately, we could take the scipy approach and have an "Expected" method on the distributions that have special cases, but I think this is cleaner.

dist/expected.go
kortschak: Why do you expect this to be a particular concern? f needs to not mutate concurrently and so does q.Quantile. I don't think either of those are particularly onerous provided that that condition is specified in the docs here.

btracey: f does not need to mutate concurrently, but frequently f will be a closure around some more complicated code. The easiest case for many uses is to be stateful (especially since many types are stateful and don't provide a "clone" function)

jonlawlor: If you are considering a boolean for `concurrent` I would suggest having 2 functions, 1 for concurrent use and 1 for synchronous.

btracey: Why? It's pretty easy for the user to decide "can my function run concurrently", and it's easier to change if the user invests the effort to make the function stateless.

jonlawlor: Same reason that the standard lib's sort has both Sort and SortStable instead of a flag on Sort.  It is easier to understand a function's name than a bool that changes a function's behavior, particularly if it is subtle.  I'm not sure it is easier or harder to change a flag or a function's name, seems about the same to me.

btracey: You're right that the easy to change argument is useless.

I don't think it's the same as Sort and SortStable. The sort code uses two different algorithms. This is exactly the same behavior, and will give exactly the same answer (up to rounding), it's only a matter if f is concurrently executable or not. For example, quadrature with a fixed number of points and quadrature with an adaptive number of points should be different functions because the algorithm is different.

jonlawlor: The concurrent version uses subintegrals though?  I feel like I am arguing semantics a bit, but it would result in slightly different output due to rounding, so to me that is a different algorithm.

btracey: The concurrent version evaluates the same locations. The number of quadrature points determines a fixed set of locations and weights. The same locations are evaluated regardless of the concurrency, the concurrency just allows several elements of that set to be evaluated simultaneously. In the end it's computing the same \sum_i w_i f(x_i) -- concurrency does not effect either the w_i or the x_i.

jonlawlor: The subintegrals can produce different rounding.  You might have values [A, B, C, D], and with the sync version you would perform A/4 + B/4 + C/4 + D/4, but in the concurrent version (with 2 values) you perform (A/4 + B/4) + (C/4 + D/4).  Either way would have some rounding, but not the same.  Take  [1e30, 1e30, -2e30, 2.0] - the first method will produce an answer of .5, the second will produce an answer of 0.0.

The thing that would worry me is that the output would depend on the scheduling of goroutines, so different evaluations of the same ExpectedFixed(...) with concurrency would not always be the same if there were more than 2 subintegrals.

btracey: Your different addition schemes I was classifying as rounding errors, since they give the same answer in infinite precision arithmetic. While I sympathize with your point, I think you're being overly picky. There are lots of cases in gonum where we change minor things -- parentheses around additions, changing native.Daxpy to use SIMD, etc. For "normal" inputs those changes don't matter, they change by a few ULPs. There are definitely cases where there are inputs like you mention, but those inputs need to be using specialized algorithms. In the particular case of quadrature, if the function values at your four quadrature points are  [1e30, 1e30, -2e30, 2.0], the estimate of the integral is completely useless no matter if 0 or 0.5 is returned. More generally, I don't see how we can possibly update code if we have to guarantee an exact double precision match for all inputs for every change.

If this kind of thing is actually a real problem for quadrature routines, we can fix the algorithm so that it does return a ULP identical answer for concurrent and non-concurrent. Right now we use a mutex and add w_i f(x_i), but we could alternately save the w_i and f(x_i) results into a slice and perform a weighted sum once they're all computed.

jonlawlor: Yes I am being very picky, and probably too picky.  I'd rather say something about an issue (that is hopefully not completely stupid) and have it get shot down than not mention it at all.  I think there might be some middle grounds where the result is the same with same concurrent input, but it isn't obvious how much a performance hit there would be.  Anyway, as long as you're ok with it then I'm fine, although you might want a note that mentions that the output won't be exactly equal if concurrent >= 2.

btracey: The discussion is definitely healthy. I do think you're right that fundamentally different algorithms deserve names, but I also think that if two things give the same answer in infinite precision they are the same algorithm. I don't think we can guarantee exact matching for floating point values always, and instead need to do what's reasonable. In the specific case, it's not so clear. What is the relative cost of 2*n float64 memory versus one mutex and locking?

I'm not sure we want that note. I agree that it's true, but it seems that it should be assumed in general that concurrency leads to non-determinstic ordering (unless stated explicitly otherwise), and that non-determinism in floating point addition leads to results that can differ slightly. In other words, the behavior you mention should be the default assumption for floating point numbers and concurrency.

btracey:
Let's leave this for the moment. I think I have a better way of doing this by transforming the quad signatures. I have a busy two weeks and can't get a PR to submission quality before then.

PR #92, opened 2016-03-05
stat: trim wide backing vecs matrix during PCA

kortschak:
vecs is wide when a is wide, but vectors beyond n are not valid, so
clone the result view into vecs. This costs an allocation when it
happens, but potentially saves significant space - when n << d.

@btracey Please take a look.


kortschak:
The code is nicer now as well, as a bonus.

btracey:
LGTM.

PR #96, opened 2016-03-16
ROC function

Armadilloa16:
Produces a ROC curve.

I'm pretty new to go, and to software development in general
so I'm in the middle of a pretty steep learning curve -- I thought
I should start with some small, simple contributions as I learn.
Any feedback would be very helpful.

@kortschak  @btracey please take a look?

### DIFF
--- a/roc.go
+++ b/roc.go
@@ -1,0 +1,121 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package stat
+
+import "sort"
+
+// ROC returns paired false positive rate (FPR) and true positive rate
+// (TPR) values corresponding to n cutoffs spanning the relative
+// (or receiver) operator characteristic (ROC) curve obtained when y is
+// treated as a binary classifier for classes with weights.
+//
+// Cutoffs are equally spaced from eps less than the minimum value of y
+// to the maximum value of y, including both endpoints meaning that the
+// resulting ROC curve will always begin at (0,0) and end at (1,1).
+//
+// The input y must be sorted, and SortWeightedLabeled can be used in
+// order to sort y together with classes and weights.
+//
+// For a given cutoff value, observations corresponding to entries in y
+// greater than the cutoff value are classified as false, while those
+// below (or equal to) the cutoff value are classified as true. These
+// assigned class labels are compared with the true values in the classes
+// slice and used to calculate the FPR and TPR.
+//
+// If weights is nil, all weights are treated as 1.
+//
+// When n is zero all possible cutoffs are calculated, resulting
+// in fpr and tpr having length one greater than the number of unique
+// values in y. When n is greater than one fpr and tpr will be returned
+// with length n. ROC will panic if n is equal to one or less than 0.
+//
+// More details about ROC curves are available at
+// https://en.wikipedia.org/wiki/Receiver_operating_characteristic
+func ROC(n int, y []float64, classes []bool, weights []float64) (tpr, fpr []float64) {
+	if len(y) != len(classes) {
+		panic("stat: slice length mismatch")
+	}
+	if weights != nil && len(y) != len(weights) {
+		panic("stat: slice length mismatch")
+	}
+	if !sort.Float64sAreSorted(y) {
+		panic("stat: input must be sorted")
+	}
+
+	var incWidth, tol float64
+	if n == 0 {
+		if len(y) == 0 {
+			return nil, nil
+		}
+		tpr = make([]float64, len(y)+1)
+		fpr = make([]float64, len(y)+1)
+	} else {
+		if n < 2 {
+			panic("stat: cannot calculate fewer than 2 points on a ROC curve")
+		}
+		if len(y) == 0 {
+			return nil, nil
+		}
+		tpr = make([]float64, n)
+		fpr = make([]float64, n)
+		incWidth = (y[len(y)-1] - y[0]) / float64(n-1)
+		tol = y[0] + incWidth
+		if incWidth == 0 {
+			tpr[n-1] = 1
+			fpr[n-1] = 1
+			return
+		}
+	}
+
+	var bin int = 1 // the initial bin is known to have 0 fpr and 0 tpr
+	var nPos, nNeg float64
+	for i, u := range classes {
+		var posWeight, negWeight float64 = 0, 1
+		if weights != nil {
+			negWeight = weights[i]
+		}
+		if u {
+			posWeight, negWeight = negWeight, posWeight
+		}
+		nPos += posWeight
+		nNeg += negWeight
+		tpr[bin] += posWeight
+		fpr[bin] += negWeight
+
+		// Assess if the bin needs to be updated. If n is zero,
+		// the bin is always updated, unless consecutive y values
+		// are equal. Otherwise, the bin must be updated until it
+		// matches the next y value (skipping empty bins).
+		if n == 0 {
+			if i != (len(y)-1) && y[i] != y[i+1] {
+				bin++
+				tpr[bin] = tpr[bin-1]
+				fpr[bin] = fpr[bin-1]
+			}
+		} else {
+			for i != (len(y)-1) && y[i+1] > tol {
+				tol += incWidth
+				bin++
+				tpr[bin] = tpr[bin-1]
+				fpr[bin] = fpr[bin-1]
+			}
+		}
+	}
+	if n == 0 {
+		tpr = tpr[:(bin + 1)]
+		fpr = fpr[:(bin + 1)]
+	}
+
+	invNeg := 1 / nNeg
+	invPos := 1 / nPos
+	for i := range tpr {
+		tpr[i] *= invPos
+		fpr[i] *= invNeg
+	}
+	tpr[len(tpr)-1] = 1
+	fpr[len(fpr)-1] = 1
+
+	return tpr, fpr
+}
--- a/roc_test.go
+++ b/roc_test.go
@@ -1,0 +1,178 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package stat
+
+import (
+	"testing"
+
+	"github.com/gonum/floats"
+)
+
+// Test cases where calculated manually.
+func TestROC(t *testing.T) {
+	cases := []struct {
+		y       []float64
+		c       []bool
+		w       []float64
+		n       int
+		wantTPR []float64
+		wantFPR []float64
+	}{
+		{
+			y:       []float64{0, 3, 5, 6, 7.5, 8},
+			c:       []bool{true, false, true, false, false, false},
+			wantTPR: []float64{0, 0.5, 0.5, 1, 1, 1, 1},
+			wantFPR: []float64{0, 0, 0.25, 0.25, 0.5, 0.75, 1},
+		},
+		{
+			y:       []float64{0, 3, 5, 6, 7.5, 8},
+			c:       []bool{true, false, true, false, false, false},
+			w:       []float64{4, 1, 6, 3, 2, 2},
+			wantTPR: []float64{0, 0.4, 0.4, 1, 1, 1, 1},
+			wantFPR: []float64{0, 0, 0.125, 0.125, 0.5, 0.75, 1},
+		},
+		{
+			y:       []float64{0, 3, 5, 6, 7.5, 8},
+			c:       []bool{true, false, true, false, false, false},
+			n:       int(5),
+			wantTPR: []float64{0, 0.5, 0.5, 1, 1},
+			wantFPR: []float64{0, 0, 0.25, 0.5, 1},
+		},
+		{
+			y:       []float64{0, 3, 5, 6, 7.5, 8},
+			c:       []bool{true, false, true, false, false, false},
+			n:       int(9),
+			wantTPR: []float64{0, 0.5, 0.5, 0.5, 0.5, 1, 1, 1, 1},
+			wantFPR: []float64{0, 0, 0, 0.25, 0.25, 0.25, 0.5, 0.5, 1},
+		},
+		{
+			y:       []float64{0, 3, 5, 6, 7.5, 8},
+			c:       []bool{true, false, true, false, false, false},
+			w:       []float64{4, 1, 6, 3, 2, 2},
+			n:       int(5),
+			wantTPR: []float64{0, 0.4, 0.4, 1, 1},
+			wantFPR: []float64{0, 0, 0.125, 0.5, 1},
+		},
+		{
+			y:       []float64{0, 3, 5, 6, 7.5, 8},
+			c:       []bool{true, false, true, false, false, false},
+			w:       []float64{4, 1, 6, 3, 2, 2},
+			n:       int(9),
+			wantTPR: []float64{0, 0.4, 0.4, 0.4, 0.4, 1, 1, 1, 1},
+			wantFPR: []float64{0, 0, 0, 0.125, 0.125, 0.125, 0.5, 0.5, 1},
+		},
+		{
+			y:       []float64{0, 3, 6, 6, 6, 8},
+			c:       []bool{true, false, true, false, false, false},
+			wantTPR: []float64{0, 0.5, 0.5, 1, 1},
+			wantFPR: []float64{0, 0, 0.25, 0.75, 1},
+		},
+		{
+			y:       []float64{0, 3, 6, 6, 6, 8},
+			c:       []bool{true, false, true, false, false, false},
+			w:       []float64{4, 1, 6, 3, 2, 2},
+			wantTPR: []float64{0, 0.4, 0.4, 1, 1},
+			wantFPR: []float64{0, 0, 0.125, 0.75, 1},
+		},
+		{
+			y:       []float64{0, 3, 6, 6, 6, 8},
+			c:       []bool{true, false, true, false, false, false},
+			n:       int(5),
+			wantTPR: []float64{0, 0.5, 0.5, 1, 1},
+			wantFPR: []float64{0, 0, 0.25, 0.75, 1},
+		},
+		{
+			y:       []float64{0, 3, 6, 6, 6, 8},
+			c:       []bool{true, false, true, false, false, false},
+			n:       int(9),
+			wantTPR: []float64{0, 0.5, 0.5, 0.5, 0.5, 0.5, 1, 1, 1},
+			wantFPR: []float64{0, 0, 0, 0.25, 0.25, 0.25, 0.75, 0.75, 1},
+		},
+		{
+			y:       []float64{0, 3, 6, 6, 6, 8},
+			c:       []bool{true, false, true, false, false, false},
+			w:       []float64{4, 1, 6, 3, 2, 2},
+			n:       int(5),
+			wantTPR: []float64{0, 0.4, 0.4, 1, 1},
+			wantFPR: []float64{0, 0, 0.125, 0.75, 1},
+		},
+		{
+			y:       []float64{0, 3, 6, 6, 6, 8},
+			c:       []bool{true, false, true, false, false, false},
+			w:       []float64{4, 1, 6, 3, 2, 2},
+			n:       int(9),
+			wantTPR: []float64{0, 0.4, 0.4, 0.4, 0.4, 0.4, 1, 1, 1},
+			wantFPR: []float64{0, 0, 0, 0.125, 0.125, 0.125, 0.75, 0.75, 1},
+		},
+		{
+			y:       []float64{1, 2},
+			c:       []bool{true, true},
+			wantTPR: []float64{0, 0.5, 1},
+			wantFPR: []float64{0, 0, 1},
+		},
+		{
+			y:       []float64{1, 2},
+			c:       []bool{true, true},
+			n:       int(2),
+			wantTPR: []float64{0, 1},
+			wantFPR: []float64{0, 1},
+		},
+		{
+			y:       []float64{1, 2},
+			c:       []bool{true, true},
+			n:       int(7),
+			wantTPR: []float64{0, 0.5, 0.5, 0.5, 0.5, 0.5, 1},
+			wantFPR: []float64{0, 0, 0, 0, 0, 0, 1},
+		},
+		{
+			y:       []float64{1},
+			c:       []bool{true},
+			wantTPR: []float64{0, 1},
+			wantFPR: []float64{0, 1},
+		},
+		{
+			y:       []float64{1},
+			c:       []bool{true},
+			n:       int(2),
+			wantTPR: []float64{0, 1},
+			wantFPR: []float64{0, 1},
+		},
+		{
+			y:       []float64{1},
+			c:       []bool{false},
+			wantTPR: []float64{0, 1},
+			wantFPR: []float64{0, 1},
+		},
+		{
+			y:       []float64{0.01, 0.02, 0.03, 0.04, 0.05, 0.06, 10},
+			c:       []bool{true, false, true, true, false, false, true},
+			n:       int(5),
+			wantTPR: []float64{0, 0.75, 0.75, 0.75, 1},
+			wantFPR: []float64{0, 1, 1, 1, 1},
+		},
+		{
+			y:       []float64{},
+			c:       []bool{},
+			wantTPR: nil,
+			wantFPR: nil,
+		},
+		{
+			y:       []float64{},
+			c:       []bool{},
+			n:       int(5),
+			wantTPR: nil,
+			wantFPR: nil,
+		},
+	}
+	for i, test := range cases {
+		gotTPR, gotFPR := ROC(test.n, test.y, test.c, test.w)
+		if !floats.Same(gotTPR, test.wantTPR) {
+			t.Errorf("%d: unexpected TPR got:%v want:%v", i, gotTPR, test.wantTPR)
+		}
+		if !floats.Same(gotFPR, test.wantFPR) {
+			t.Errorf("%d: unexpected FPR got:%v want:%v", i, gotFPR, test.wantFPR)
+		}
+	}
+}
--- a/stat.go
+++ b/stat.go
@@ -864,17 +864,66 @@ type weightSorter struct {
 	w []float64
 }
 
-func (w weightSorter) Less(i, j int) bool {
-	return w.x[i] < w.x[j]
-}
-
+func (w weightSorter) Len() int           { return len(w.x) }
+func (w weightSorter) Less(i, j int) bool { return w.x[i] < w.x[j] }
 func (w weightSorter) Swap(i, j int) {
 	w.x[i], w.x[j] = w.x[j], w.x[i]
 	w.w[i], w.w[j] = w.w[j], w.w[i]
 }
 
-func (w weightSorter) Len() int {
-	return len(w.x)
+// SortWeightedLabeled rearranges the data in x along with their
+// corresponding weights and boolean labels so that the x data are sorted.
+// The data is sorted in place. Weights and labels may be nil, if either
+// is non-nil it must have the same length as x.
+func SortWeightedLabeled(x []float64, labels []bool, weights []float64) {
+	if labels == nil {
+		SortWeighted(x, weights)
+		return
+	}
+	if weights == nil {
+		if len(x) != len(labels) {
+			panic("stat: slice length mismatch")
+		}
+		sort.Sort(labelSorter{
+			x: x,
+			l: labels,
+		})
+		return
+	}
+	if len(x) != len(labels) || len(x) != len(weights) {
+		panic("stat: slice length mismatch")
+	}
+	sort.Sort(weightLabelSorter{
+		x: x,
+		l: labels,
+		w: weights,
+	})
+}
+
+type labelSorter struct {
+	x []float64
+	l []bool
+}
+
+func (a labelSorter) Len() int           { return len(a.x) }
+func (a labelSorter) Less(i, j int) bool { return a.x[i] < a.x[j] }
+func (a labelSorter) Swap(i, j int) {
+	a.x[i], a.x[j] = a.x[j], a.x[i]
+	a.l[i], a.l[j] = a.l[j], a.l[i]
+}
+
+type weightLabelSorter struct {
+	x []float64
+	l []bool
+	w []float64
+}
+
+func (a weightLabelSorter) Len() int           { return len(a.x) }
+func (a weightLabelSorter) Less(i, j int) bool { return a.x[i] < a.x[j] }
+func (a weightLabelSorter) Swap(i, j int) {
+	a.x[i], a.x[j] = a.x[j], a.x[i]
+	a.l[i], a.l[j] = a.l[j], a.l[i]
+	a.w[i], a.w[j] = a.w[j], a.w[i]
 }
 
 // StdDev returns the sample standard deviation.
--- a/stat_test.go
+++ b/stat_test.go
@@ -7,6 +7,7 @@ package stat
 import (
 	"fmt"
 	"math"
+	"reflect"
 	"testing"
 
 	"github.com/gonum/floats"
@@ -1260,6 +1261,65 @@ func TestSortWeighted(t *testing.T) {
 		}
 	}
 	if !Panics(func() { SortWeighted(make([]float64, 3), make([]float64, 2)) }) {
+		t.Errorf("SortWeighted did not panic with x, weights length mismatch")
+	}
+}
+
+func TestSortWeightedLabeled(t *testing.T) {
+	for i, test := range []struct {
+		x    []float64
+		l    []bool
+		w    []float64
+		ansx []float64
+		ansl []bool
+		answ []float64
+	}{
+		{
+			x:    []float64{8, 3, 7, 8, 4},
+			ansx: []float64{3, 4, 7, 8, 8},
+		},
+		{
+			x:    []float64{8, 3, 7, 8, 4},
+			w:    []float64{.5, 1, 1, .5, 1},
+			ansx: []float64{3, 4, 7, 8, 8},
+			answ: []float64{1, 1, 1, .5, .5},
+		},
+		{
+			x:    []float64{8, 3, 7, 8, 4},
+			l:    []bool{false, false, true, false, true},
+			ansx: []float64{3, 4, 7, 8, 8},
+			ansl: []bool{false, true, true, false, false},
+		},
+		{
+			x:    []float64{8, 3, 7, 8, 4},
+			l:    []bool{false, false, true, false, true},
+			w:    []float64{.5, 1, 1, .5, 1},
+			ansx: []float64{3, 4, 7, 8, 8},
+			ansl: []bool{false, true, true, false, false},
+			answ: []float64{1, 1, 1, .5, .5},
+		},
+	} {
+		SortWeightedLabeled(test.x, test.l, test.w)
+		if !floats.Same(test.x, test.ansx) {
+			t.Errorf("SortWeightedLabelled mismatch case %d. Expected x %v, Found x %v", i, test.ansx, test.x)
+		}
+		if (test.l != nil) && !reflect.DeepEqual(test.l, test.ansl) {
+			t.Errorf("SortWeightedLabelled mismatch case %d. Expected l %v, Found l %v", i, test.ansl, test.l)
+		}
+		if (test.w != nil) && !floats.Same(test.w, test.answ) {
+			t.Errorf("SortWeightedLabelled mismatch case %d. Expected w %v, Found w %v", i, test.answ, test.w)
+		}
+	}
+	if !Panics(func() { SortWeightedLabeled(make([]float64, 3), make([]bool, 2), make([]float64, 3)) }) {
+		t.Errorf("SortWeighted did not panic with x, labels length mismatch")
+	}
+	if !Panics(func() { SortWeightedLabeled(make([]float64, 3), make([]bool, 2), nil) }) {
+		t.Errorf("SortWeighted did not panic with x, labels length mismatch")
+	}
+	if !Panics(func() { SortWeightedLabeled(make([]float64, 3), make([]bool, 3), make([]float64, 2)) }) {
+		t.Errorf("SortWeighted did not panic with x, weights length mismatch")
+	}
+	if !Panics(func() { SortWeightedLabeled(make([]float64, 3), nil, make([]float64, 2)) }) {
 		t.Errorf("SortWeighted did not panic with x, weights length mismatch")
 	}
 }


Armadilloa16:
Whoops should probably fix that first.

roc.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (a ByX) Len() int { return len(a.x) }
+func (a ByX) Swap(i, j int) {
+	a.x[i], a.x[j] = a.x[j], a.x[i]
+	a.c[i], a.c[j] = a.c[j], a.c[i]
+}
+func (a ByX) Less(i, j int) bool { return a.x[i] < a.x[j] }
+
+
+// ROC returns paired FPR and TPR values which describe the
```
sbinet: you may want to mention that `ROC` will modify `y` and `pos`

roc.go
```diff
@@ -0,0 +1,53 @@
...
+}
+func (a ByX) Less(i, j int) bool { return a.x[i] < a.x[j] }
+
+
+// ROC returns paired FPR and TPR values which describe the 
+// ROC curve treating y as a binary classifier for pos.
+func ROC(y []float64, pos []bool) ([]float64, []float64) {
+
+	if len(y) != len(pos) {
+		panic("inputs differ in length")
```
sbinet: please prepend the package name _e.g._
```go
panic("stat: length mismatch")
```

roc.go
```diff
@@ -0,0 +1,53 @@
...
+func (a ByX) Len() int { return len(a.x) }
+func (a ByX) Swap(i, j int) {
+	a.x[i], a.x[j] = a.x[j], a.x[i]
+	a.c[i], a.c[j] = a.c[j], a.c[i]
+}
+func (a ByX) Less(i, j int) bool { return a.x[i] < a.x[j] }
+
+
+// ROC returns paired FPR and TPR values which describe the 
+// ROC curve treating y as a binary classifier for pos.
```
sbinet: also perhaps explain a bit what ROC (receiver operating characteristic), TPR (true positive rate) and FPR (false positive rate) stand for.
you couls also point to wikipedia for some more detailed informations.

roc.go
```diff
@@ -0,0 +1,53 @@
...
+func ROC(y []float64, pos []bool) ([]float64, []float64) {
```
sbinet: perhaps modify the signature of `ROC` to:

```go
func ROC(y []float64, pos []bool) (tpr, fpr []float64) { ... }
```
so we know who's who.

roc_test.go
```diff
@@ -0,0 +1,53 @@
...
+// license that can be found in the LICENSE file.
+
+package stat
+
+import (
+	"math/rand"
+	"testing"
+)
+
+func TestROC(t *testing.T) {
```
sbinet: please explain how the test cases' values were obtained (_e.g._ using `numpy`+`scipy` with such and such data and such and such functions)

sbinet:
ah, the build fails because `roc.go` has not been `gofmt`ed.
you can do it like so:

```sh
$> gofmt -w .
```

roc.go
btracey: "stat: input length mismatch"

roc.go
```diff
@@ -0,0 +1,53 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package stat
+
+import "sort"
+
+type ByX struct {
```
btracey: This shouldn't be exported.

btracey:
Is there a concise description of what the ROC curve does? Wikipedia says that it's a plot of true positive rate and true negative rate as a function of tolerance. With that discription, I would have expected the signature to be something like
````
     ROC(tolerenaces, obs []float64, ans []bool) (fpr, tnr) 
````

Where tolerance is the set of values for which we want to know the ROC, obs is the set of observations on which we make the decision, and ans is if it was actually a positive.

I don't understand how that maps to your function.

File: roc.go [hunk 7]
```diff
@@ -0,0 +1,53 @@
...
+	a.x[i], a.x[j] = a.x[j], a.x[i]
+	a.c[i], a.c[j] = a.c[j], a.c[i]
+}
+func (a ByX) Less(i, j int) bool { return a.x[i] < a.x[j] }
+
+
+// ROC returns paired FPR and TPR values which describe the 
+// ROC curve treating y as a binary classifier for pos.
+func ROC(y []float64, pos []bool) ([]float64, []float64) {
+
```
kortschak: Delete blank line,

roc.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	sort.Sort(ByX{y, pos})
+
+	var np float64
+	tpr := make([]float64, len(y)+1)
+	fpr := make([]float64, len(y)+1)
+	for i, u := range pos {
+		if u {
+			np += 1.0
```
kortschak: np++

roc.go
```diff
@@ -0,0 +1,53 @@
...
+			tpr[i+1] = tpr[i] + 1.0
+			fpr[i+1] = fpr[i]
+		} else {
+			tpr[i+1] = tpr[i]
+			fpr[i+1] = fpr[i] + 1.0
+		}
+	}
+	nn := float64(len(y)) - np
+	for i, _ := range tpr {
```
kortschak: for i := range tpr

roc.go
```diff
@@ -0,0 +1,53 @@
...
+		if u {
+			np += 1.0
+			tpr[i+1] = tpr[i] + 1.0
+			fpr[i+1] = fpr[i]
+		} else {
+			tpr[i+1] = tpr[i]
+			fpr[i+1] = fpr[i] + 1.0
+		}
+	}
+	nn := float64(len(y)) - np
```
kortschak: `nn := 1/(float(len(y)) - np)` and at L48 `fpr[i] *= nn`
`np = 1/np` and at L47 `tpr[i] *= np`.

roc_test.go
```diff
@@ -0,0 +1,53 @@
...
+
```
kortschak: Add `rnd := rand.New(rand.NewSource(1))` and use `rnd.Perm` instead of `rand.Perm`.

roc_test.go
```diff
@@ -0,0 +1,53 @@
...
+			tpr: []float64{0.0, 0.25, 0.5, 0.75, 0.75, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0},
+			fpr: []float64{0.0, 0.0, 0.0, 0.0, 0.125, 0.125, 0.25, 0.375, 0.5, 0.625, 0.75, 0.875, 1.0},
+		},
+	}
+	p := rand.Perm(len(cases[0].y))
+	for i, j := range p {
+		cases[1].y[i] = cases[0].y[j]
+		cases[1].pos[i] = cases[0].pos[j]
+	}
+	for _, c := range cases {
```
kortschak: s/c/test/g

roc_test.go
```diff
@@ -0,0 +1,53 @@
...
+		tpr, fpr := ROC(c.y, c.pos)
+		if len(tpr) != len(fpr) {
+			t.Fatalf("len(tpr) (%d) does not equal len(fpr) (%d)", len(tpr), len(fpr))
+		}
+		if len(tpr) != len(c.tpr) {
+			t.Fatalf("len(tpr) (%d) does not equal expected length (%d)", len(tpr), len(c.tpr))
+		}
+		for i, _ := range tpr {
```
kortschak: Replace for loop (should omit `, _` BTW) with:

```
if !reflect.DeepEqual(gotTpr, c.wantTpr) {
    t.Errorf("unexpected TPR got:%v want:%v", gotTpr, test.wantTpr)
}
if !reflect.DeepEqual(gotTfr, c.wantTfr) {
    t.Errorf("unexpected TFR got:%v want:%v", gotTfr, test.wantTfr)
}
```

If you add `i` to the cases for loop, you can report which test failed here too.

roc_test.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+	}
+	p := rand.Perm(len(cases[0].y))
+	for i, j := range p {
+		cases[1].y[i] = cases[0].y[j]
+		cases[1].pos[i] = cases[0].pos[j]
+	}
+	for _, c := range cases {
+		tpr, fpr := ROC(c.y, c.pos)
+		if len(tpr) != len(fpr) {
```
kortschak: If reflect.DeepEqual is used, this is redundant.

(must import reflect)

kortschak:
@btracey the tolerances are implicit in the returned slices.

roc.go
Armadilloa16: I'm not sure what this means -- I feel like I am missing something super simple :/

roc.go
Armadilloa16: Yep did both.

roc.go
Armadilloa16: done.

roc_test.go
Armadilloa16: Uhhh... actually I just made them up on the fly (in my head). 
I should probably check them somewhere else I suppose?

roc.go
Armadilloa16: Fixed.

roc_test.go
Armadilloa16: Not sure what this means? Am I missing something simple again?

roc_test.go
Armadilloa16: Done.

roc_test.go
Armadilloa16: Yep.

Armadilloa16:
> Is there a concise description of what the ROC curve does? Wikipedia says that it's a plot of true positive rate and true negative rate as a function of tolerance. With that discription, I would have expected the signature to be something like
> `     ROC(tolerenaces, obs []float64, ans []bool) (fpr, tnr) `
> Where tolerance is the set of values for which we want to know the ROC, obs is the set of observations on which we make the decision, and ans is if it was actually a positive.
> 
> I don't understand how that maps to your function.

@btracey I've added a concise(ish) description of what a ROC curve does. Also, you 
are right in the sense that you could make a function to take tolerances (or cutoffs 
whichever term you prefer), what a ROC curve (and this function) does is essentially
calculate the fpr and tpr's for _all_ possible cutoffs. These will only change at the 
values of y, and so assuming the values of y are distinct the output will satisfy
len(fpr) == len(y) + 1, with each value corresponding to the value of fpr/tpr for any 
tolerance in the corresponding interval -- there are len(y) + 1 distinct intervals, the 
first corresponding to any tolerance less than the minimum value in y, the last 
corresponding to any tolerance greater than the maximum value in y.

Thanks for all the constructive comments everyone! Super helpful, I am learning so 
much! :D

kortschak:
What happens when the values of y are not distinct?

roc.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (a byX) Len() int { return len(a.x) }
+func (a byX) Swap(i, j int) {
+	a.x[i], a.x[j] = a.x[j], a.x[i]
+	a.c[i], a.c[j] = a.c[j], a.c[i]
+}
+func (a byX) Less(i, j int) bool { return a.x[i] < a.x[j] }
+
+// ROC sorts both inputs for increasing values in y, and
```
kortschak: Normally we start doc comments along the lines "ROC returns the ..." Would you prefix this?

Armadilloa16:
Done. PTAL

@kortschak Good question haha. Now I think about it it is actually a 
problem. 

So long as each set of duplicate values in y correspond to a single 
value in pos, its all good. For plotting purposes and calculating area 
under curves (AUC), etc. ROC will return some redundant values. 
If you consider some of the len(y)+1 intervals I mentioned above to 
be width zero then this is still correct, will still produce correct plots 
and AUC curves, and produce consistent output, so I think this 
behaviour is acceptable -- the only downside I see is that it makes the 
cutoff interpretation a little funky.

HOWEVER, if at least two equal values in y correspond to different values 
of pos, then what is `correct' is actually ambiguous, and ROC will simply 
return one of the possibilities. The only way I can think of to address this 
would be to remove values in fpr and tpr corresponding to intervals 
between duplicate values in y. This would result in diagonal lines in the ROC 
curve, which would be the only solution I could think of that would be 
unique. Seems to me like the best solution. However it would also significantly 
complicate the code. Worth doing? An alternative solution is to leave the function
as it is, and note that we assume y consists of distinct values (or at least that 
duplicate values correspond to single values in pos). What do you think?

kortschak:
What does R do in this situation?

roc.go
```diff
@@ -0,0 +1,62 @@
...
+// For a given cutoff value, observations corresponding to
+// entries in y greater than the cutoff are classified as
+// true, and these assigned class labels are compared to the
+// true class labels in pos to calculate FPR and TPR for all
+// possible cutoffs.
+// More details about ROC curves are available at
+// https://en.wikipedia.org/wiki/Receiver_operating_characteristic
+// Note that ROC will modify both inputs -- sorting them in order
+// of increasing y values.
+func ROC(y []float64, pos []bool) (tpr, fpr []float64) {
```
kortschak: You use y for the label here, but byX sorts by the field x. Either change y or x.

Armadilloa16:
r-base does not have a ROC function. And the packages with that functionality are pretty dodge in my experience. 

This guy:
http://www.r-bloggers.com/illustrated-guide-to-roc-and-auc/
avoids the issue by simply calculating for a fixed number of 
equally spaced thresholds, similar to what @btracey suggested.

This package:
https://cran.r-project.org/web/packages/pROC/index.html
does the removing between-duplicates intervals thing I mentioned
when plotting.

As does this package:
https://rocr.bioinf.mpi-sb.mpg.de/

So although I don't think any of those packages are actually any 
good (honestly), I do think that the idea of removing between-duplicate
interval values is a good idea and I'll have a go at implementing it 
tomorrow, despite the fact that it will complicate the code abit. I'll 
also add a test for those edge cases.

roc.go
```diff
@@ -6,35 +6,44 @@ package stat
...
-
-// ROC returns paired FPR and TPR values which describe the 
-// ROC curve treating y as a binary classifier for pos.
-func ROC(y []float64, pos []bool) ([]float64, []float64) {
-
+func (a byX) Less(i, j int) bool { return a.x[i] < a.x[j] }
+
+// ROC returns paired false positive rate (FPR) and true positive
+// rate (TPR) values which describe the relative (or receiver)
+// operator chracteristic (ROC) curve obtained when y is
```
sbinet: typo:
s/chracteristic/characteristic/

roc.go
Armadilloa16: Done. Also changed a bunch of other variable names to be more concise/ prettier.
Is it ok to use gotFPR instead of the perhaps more usual gotFpr?

kortschak: gotFPR (FPR is an initialism, not a word).

roc_test.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+	p := rnd.Perm(len(cases[0].y))
+	for i, j := range p {
+		cases[1].y[i] = cases[0].y[j]
+		cases[1].c[i] = cases[0].c[j]
+	}
+	for _, test := range cases {
+		gotTPR, gotFPR := ROC(test.y, test.c)
+		if !reflect.DeepEqual(gotTPR, test.wantTPR) {
+			t.Errorf("unexpected TPR got:%v want:%v", gotTPR, test.wantTPR)
```
sbinet: please replace:
```go
t.Errorf("unexpected TPR got:%v want:%v", gotTPR, test.wantTPR)
```
with:
```go
t.Errorf("test #%d: unexpected TPR got:%v want:%v", i, gotTPR, test.wantTPR)
```
(or something along these lines)

where `i` is coming from:
```go
for i, test := range cases {
```

this way we know which testcase failed.

roc_test.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+		if !reflect.DeepEqual(gotFPR, test.wantFPR) {
+			t.Errorf("unexpected TFR got:%v want:%v", gotFPR, test.wantFPR)
```
sbinet: ditto: please add the test case number in the error message.

Armadilloa16:
Implemented the removing between-duplicate interval values thing.
Still need to add cases that test that edge case though, I'll do that tomorrow morning 
time to go rockclimbing now. :) I've had alot of fun working on this today thanks everyone 
for all your comments its been really helpful, and I've learnt a ton already. :)

roc.go
```diff
@@ -0,0 +1,71 @@
...
+	sort.Sort(byY{y, c})
+
+	var j int
+	var np float64
+	tpr = make([]float64, len(y)+1)
+	fpr = make([]float64, len(y)+1)
+	for i, u := range c {
+		if u {
+			np++
+			tpr[j+1] = tpr[j] + 1.0
```
kortschak: s/1.0/1/ and below.

btracey:
Wouldn't it be better to have both tolerances and ys? If we have 
````
ROC(tolerenaces, obs []float64, ans []bool) (fpr, tnr []float64) 
````
 pretty easy to write
````
fpr, tnr := ROC(y,y, values)
````
but if there are a lot of y values a smaller tolerances slice may be way easier to work with.

File: roc.go [hunk 21]
```diff
@@ -0,0 +1,71 @@
...
+// For a given cutoff value, observations corresponding to
+// entries in y greater than the cutoff are classified as
+// true, and these assigned class labels are compared to the
+// true class labels in c to calculate FPR and TPR for all
+// possible cutoffs.
+// More details about ROC curves are available at
+// https://en.wikipedia.org/wiki/Receiver_operating_characteristic
+// Note that ROC will modify both inputs -- sorting them in order
+// of increasing y values.
+func ROC(y []float64, c []bool) (tpr, fpr []float64) {
```
btracey: This should allow for weighted data.

btracey: Although I realize that works weird with the current definition. If my suggestion for a tolerances slice is adopted then weighted data is easy conceptually.

roc.go
```diff
@@ -0,0 +1,71 @@
...
+	if len(y) != len(c) {
+		panic("stat: inputs differ in length")
+	}
+
+	sort.Sort(byY{y, c})
```
btracey: In the other functions where data needs to be sorted (i.e. CDF) the data must be sorted at input. I think this should be the case here too.

roc.go
Armadilloa16: What do you think about making c []float64 instead of bool and specifying that positive values/ negative values indicate the classes and the values indicate weights? So using 1s and -1s would correspond to the unweighted case?

Also what do you think of having two functions -- leave this one (possibly adding weights), as this one is conveinient for plotting and also currently outputs the most concise representation for all the information in the ROC, and add a second function that takes tolerances as well. If so what would you name the functions?

One more thing to think about: we could also return cost -- so defaulting to equally weighting false positives and false negatives, but allowing for skewed weightings. I have not read it in detail (yet) but
http://link.springer.com/article/10.1007/s10994-009-5119-5
seems to make a pretty valid argument against the use of AUC etc. based on the point that 
different situations weight false positives and false negatives with varying relative importance, so returning something like the cost function could be valuable. Like I said I haven't read all the way through the above article (yet) but we could potentially return whatever alternative Hand is suggesting as well, in principle.

kortschak: >What do you think about making c []float64 instead of bool and specifying that positive values/ negative values indicate the classes and the values indicate weights? So using 1s and -1s would correspond to the unweighted case?

Please don't do this. We use a nil to indicate absence of weighting.

roc.go
kortschak: If we do that then we need to provide a way for the caller to sort the values and states together. I'm not sure that's worth it.

roc.go
kortschak: >Also what do you think of having two functions -- leave this one (possibly adding weights), as this one is conveinient for plotting and also currently outputs the most concise representation for all the information in the ROC, and add a second function that takes tolerances as well. If so what would you name the functions?

If we add tolerances then the idiom @btracey proposes works, or just nil, similar to the non-weighted case.

I *am* struggling to understand what the meaning of a weighted ROC would be though.

Armadilloa16: I'm currently leaning towards leaving it as it is, as the ROC(y,y, values) solution would produce repeated values which is ugly. There is always the option of changing it later.

I think there are some crasy cases when weighting does make sense -- say your classifying buses or something and weight each bus according to how many people are in the bus. But it requires a pretty special case for that to make sense.

sbinet:
at the risk of mudying further the water, here is what the scikit-learn toolkit does:
 http://scikit-learn.org/stable/auto_examples/model_selection/plot_roc.html

btracey:
A weighted value means the same thing it does in other places. Instead of adding a 1.0, one adds a weight[i]. 

This could come up in a number of ways. The easiest way is that values are repeated -- the same event with the same tolerance occurs. The second way is one I come up with frequently, which are that the values have unequal importance. For ROC curves, this could be time discounting due to a slightly changing environment. Samples at the moment are more important that samples from a long time ago (but the old ones aren't useless).

My suggestion about doubling up the y was wrong. Instead, the suggestion should have been that the user gets to set how many ROC values are computed. Let's say the user has a million example points. It is probably not necessary to plot a million tiny switches. Instead the user could specify to compute 1000 of them. 

ROC(n int, thresh, weights []float64, data []bool)

On the sorting, SortWeightedBool seems fine. It seems like an easier solution than forcing the user to remember which functions require sorted data and which do not. I do not like  the current implementation because it modifies the input data. As much as possible, functions should have pure inputs. If we sort the data for them we should copy it and leave the input data untouched.

kortschak:
Thanks for clarifying those. OK. Does the weighting also then gives us an approach to dealing with non-unique values?

btracey:
I don't think the weighting changes that problem. I do think specifying a certain number of ROC curve points does. Lets say all of the weights are uniform, there are a million points, and the number of ROC curve points is 1000. The function tabulates on the first 1000 points. If the 1000th point happens to equal the 1001th point, then it can be included as well. I'm not sure what should happen if all of the points have the same threshold. Probably just repeat the ending point?

Armadilloa16:
Ok I added a second function, ROCn, which allows for only n points on the ROC curve to be generated. 
I just need to write some tests for it, but does that look like roughly what you wanted?

Also, I fixed some mistakes I made earlier.

Armadilloa16:
I still have to change how the sorting is done though.

btracey:
I don't know if @kortschak agrees with me, but I would really rather that
a) It be one function
b) It not modify the inputs

On point a, it doesn't feel sufficiently different to merit two functions. It's really easy to input "ROC(len(y), y ...)". It may be a bit different than other libraries, but that's a perfect case to make an example -- "Here's the basic case that you're expecting".

On point b, code is so much easier to reason about when functions don't modify inputs (i.e. no side effects). There may be meta-data that a user forgets to sort. The vector may be used concurrently, and sorting creates a race condition. The code writer may know that ROC sorts the vector, and thus the code after depends on the sorting, while the reader of the code does not know the side-effect and so is totally confused. 

Sorry to be picky, and thanks for working with us.

Armadilloa16:
No worries, in principle I agree with you on both points, its the `how to implement' that I'm struggling with. For example, the current ROCn function generates values for equally spaced cutoffs, so ROCn(len(y),y,... ) is not the same as ROC(y,...). Even if it used cutoffs equally spaced in number of observations it would not be equivalent if the observation weights differed, and produces weird difficult to interpret results when len(y) is not nicely divisible by n. What about combining the two existing functions into one, and having a check for n == nil in which case it does ROC, and otherwise does ROCn?

About not modifying the inputs, the easiest way I can see to do this is to simply make a copy of the inputs, but this seems slow -- should I just do that? Or is it better to simply state that the inputs should be sorted, and expect the user to be able to figure that out?

kortschak:
I think there should be only one function. I'm less concerned about the input modification, but if a sort function is provided in conjunction with this that's OK.

n can't be nil, it's prefectly reasonable to state in the docs that len(y) (or zero if we want to do that - I'm fine with that) is required if the number of cut-offs should match the number of observations.

btracey:
Modifying: I'm personally of the opinion that it should behave like the other inputs. We force the users to sort before calling. This requires a SortWeightedBool type (or equivalent name). I'm personally fine with that, but there was some disagreement. We should decide before you implement a bunch of stuff that has to change.

On the n, it can't be nil. We could allow 0 to default to len(y), but it doesn't seem like it gains anything here. I agree that it's a bit unclear what to do if n doesn't divide evenly, but I'm not sure it really matters all that much as long as we try to do something reasonable. The points aren't going to be evenly spaced along the axis no matter what. In my mind, it's kind of like a histogram. For a loose cut at the ROC curve, you can say n = 10. For a fine cut, you can say n = 1000. For the whole cut, you can say n = len(y). As long as each bucket contains approximately equal weight this will make users happy (the points are never wrong, they could just be poorly spaced). If there's one y that has a huge weight, well then the normal ROC curve is going to look crazy too.

btracey:
Okay, it seems like a sort function is appropriate then. See SortWeighted for the function to extend.

btracey:
If you're confused by what I mean about the buckets, I should have some time on Tuesday or Wednesday to write up an example code.

Armadilloa16:
About n yes I realised quickly n couldnt be nil and have used zero. Note that n = 0 will produce different results to n = len(y), as n=0 will produce the complete set (the entire ROC curve, used for calculation of correct AUC values, etc.), while n=len(y) will produce a set of fpr tpr pairs corresponding to len(y) equally spaced cutoffs -- not necessarily the same thing. But I am happy with that, I think it is reasonable.

About modifying results for the time being I have just made copies of the inputs (although that seems to be failing for older versions of go?). I'll have a look into making a SortWeightedBool type tomorrow when I have abit more time. :)

Armadilloa16:
Ok I had a go at implementing a SortWeightedBool function. Is this more or less what you had in mind @btracey? I pretty much just straight out copy pasted it from stat.go although some of the grammer in the description/ function name is abit funny in this context. 

Obviously I still have to add more tests to test the new functionality, which I'll do when we've settled on the behaviour I should be testing for (now a unsorted case should fail, for example). Should we have a check for y being sorted inside ROC? Or do we leave that to the user?

kortschak:
There should be a check for sorting of y in ROC. You can use sort.IsSorted.

roc.go
```diff
@@ -0,0 +1,154 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package stat
+
+import "sort"
+
+// SortWeightedBool rearranges the data in y along with their
```
btracey: This also needs to take in weights to be SortWeightedBool :) 

Couple of bikesheds. 1) Could you make it `x`instead of `y` to match SortWeighted. 2) Could you move it to stat.go?

roc.go
```diff
@@ -0,0 +1,154 @@
...
+// corresponding boolean weights c so that the y data are sorted.
+// The data is sorted in place.
+// c may be nil, but if c is non-nil then it must have the same
```
btracey: c or weights may be nil, but if either is non-nil it must have the same length as x.

roc.go
```diff
@@ -0,0 +1,154 @@
...
+// length as y.
+func SortWeightedBool(y []float64, c []bool) {
+	if c == nil {
+		sort.Float64s(y)
```
btracey: This can then be replaced with SortWeighted(x,weights)

btracey: Rather, there's 3 cases. The both nil is this line here, the c nil is SortWeighted, and the weightsNil needs its own sorter type.

roc.go
```diff
@@ -0,0 +1,154 @@
...
+	if len(y) != len(c) {
+		panic("stat: slice length mismatch")
+	}
+	sort.Sort(byY{
+		y: y,
+		c: c,
+	})
+}
+
+type byY struct {
```
btracey: this can be `boolSorter` instead of `byY`, and can handle the weights == nil case. Then lastly there should be a `weightBoolSorter` that does all three.

Thanks!

roc.go
```diff
@@ -0,0 +1,154 @@
...
+	y []float64
+	c []bool
+}
+
+func (a byY) Len() int { return len(a.y) }
```
btracey: newline

kortschak: I'd disagree with this call. sort.Interface implementations are usually written as a stanza as done here, though also usually with the methods in alphabetical order. That could be fixed.

Armadilloa16: The Float64Slice implementation in sort goes like this:

```
// Float64Slice attaches the methods of Interface to []float64, sorting in increasing order.
type Float64Slice []float64

func (p Float64Slice) Len() int           { return len(p) }
func (p Float64Slice) Less(i, j int) bool { return p[i] < p[j] || isNaN(p[i]) && !isNaN(p[j]) }
func (p Float64Slice) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
```

Should we adopt this style? I'm not fussed either way but I do think we should be consistent.

Armadilloa16: Also maybe I should include the isNaN checks?

Armadilloa16:
Done. I've adopted the stanza formatting of the sort package,
just for the sorter types. I've not included the isNaN checks used
in the sort package though, maybe I should add them -- the 
reason I haven't is because I wanted to look up the isNaN behavior
to make sure I understand the relevance of having it there.

roc.go
kortschak: The isNaN checks make NaN sort first.

Armadilloa16: Do we want that? I'm actually not sure what ROC will do with NaN values, should probably find out.

Armadilloa16:
Added some tests for SortWeightedLabelled. Also, maybe I am missing something
but is there a reason for floats.Same being used instead of reflect.DeepEqual? Or 
was it as simple as avoiding having to import reflect?

kortschak:
reflect.DeepEqual is not good for dealing with test values that contain NaN.

stat_test.go
```diff
@@ -1264,6 +1265,65 @@ func TestSortWeighted(t *testing.T) {
...
+			ansx: []float64{3, 4, 7, 8, 8},
+			ansl: []bool{false, true, true, false, false},
+			answ: []float64{1, 1, 1, .5, .5},
+		},
+	} {
+		SortWeightedLabelled(test.x, test.l, test.w)
+		if !floats.Same(test.x, test.ansx) {
+			t.Errorf("SortWeightedLabelled mismatch case %d. Expected x %v, Found x %v", i, test.ansx, test.x)
+		}
+		if (test.l != nil) && !reflect.DeepEqual(test.l, test.ansl) {
```
Armadilloa16: Yeah that makes sense, thanks @kortschak. So am I doing the right thing using reflect.DeepEqual here as bools can't be NaN?

kortschak: Yes, this is correct.

Armadilloa16:
Ping. @kortschak @btracey @sbinet does this look alright?

roc.go
```diff
@@ -0,0 +1,135 @@
...
+// in order to sort data y together with boolean labels c and weights w.
+// For a given cutoff value, observations corresponding to
+// entries in y greater than the cutoff are classified as
+// false, and these assigned class labels are compared to the
+// class labels in c to calculate FPR and TPR.
+// w being nil results in the unweighted case.
+// n of zero results in all possible cutoffs being calculated.
+// More details about ROC curves are available at
+// https://en.wikipedia.org/wiki/Receiver_operating_characteristic
+func ROC(y []float64, c []bool, w []float64, n int) (tpr, fpr []float64) {
```
kortschak: I'd rename w => weights and c => classifier, move n to the front.

roc.go
```diff
@@ -0,0 +1,135 @@
...
+	if len(y) != len(c) {
+		panic("stat: slice length mismatch")
+	}
+	if !sort.IsSorted(sort.Float64Slice(y)) {
+		panic("stat: input must be sorted")
+	}
+
+	var j int
```
kortschak: Move j's decl to where it is used.

roc.go
```diff
@@ -0,0 +1,135 @@
...
+	var np, nn, iw, tol float64
```
kortschak: Same with these (np and nn are not used until L56).

kortschak: Rename them to nNegative and nPositive.

roc.go
```diff
@@ -0,0 +1,135 @@
...
+			panic("stat: cannot less than 2 points on a ROC curve")
+		}
+		iw = (y[len(y)-1] - y[0]) / float64(n-1)
+		if iw <= 0 {
+			panic("stat: y constant or not sorted")
+		}
+		tol = y[0] + iw
+	}
+
+	ne := true
```
kortschak: I like short names, but this is a bit short; I don't know what it means.

kortschak: If it means notEqual, rename it to equal and invert the logic. Negations are a bad idea most of the time.

roc.go
Armadilloa16: Yep, I did c => classes instead though.

roc.go
Armadilloa16: newEntry

roc.go
```diff
@@ -10,30 +10,29 @@ import "sort"
...
+	if len(y) != len(classes) {
 		panic("stat: slice length mismatch")
 	}
 	if !sort.IsSorted(sort.Float64Slice(y)) {
 		panic("stat: input must be sorted")
 	}
 
-	var j int
-	var np, nn, iw, tol float64
+	var iw, tol float64
```
Armadilloa16: @kortschak is iw clear or should I rename it 
iw => incrementWidth ?

kortschak: incWidth

roc.go
```diff
@@ -0,0 +1,136 @@
...
+
+// ROC returns paired false positive rate (FPR) and true
+// positive rate (TPR) values corresponding to n
+// cutoffs spanning the relative (or receiver)
+// operator characteristic (ROC) curve obtained when y is
+// treated as a binary classifier for classes with weights.
+// Cutoffs are equally spaced from eps less than the minimum value of y
+// to the maximum value of y, including both endpoints -- this means the
+// resulting ROC curve will always begin at (0,0) and end at (1,1).
+// input y must be sorted, and SortWeightedLabelled can be used
```
btracey: The input y

stat.go
```diff
@@ -864,17 +864,67 @@ type weightSorter struct {
...
+func (w weightSorter) Len() int           { return len(w.x) }
+func (w weightSorter) Less(i, j int) bool { return w.x[i] < w.x[j] }
 func (w weightSorter) Swap(i, j int) {
 	w.x[i], w.x[j] = w.x[j], w.x[i]
 	w.w[i], w.w[j] = w.w[j], w.w[i]
 }
 
-func (w weightSorter) Len() int {
-	return len(w.x)
+// SortWeightedLabelled rearranges the data in x along with their
```
btracey: We use american english, right? SortWeightedLabeled

roc.go
btracey: SortWeightedLabeled

roc.go
```diff
@@ -0,0 +1,136 @@
...
+// in order to sort data y together with classes and weights.
```
btracey: sort y together with

stat.go
kortschak: This is true - for client facing documentation. It is the price we pay.

btracey: Indeed. Should have taken up the reign when the empire started declining.

roc.go
```diff
@@ -0,0 +1,136 @@
...
+// For a given cutoff value, observations corresponding to
```
btracey: Aren't values above the cutoff classified as true?

This is confusing. Here's what I would say

For a given cutoff value, observations corresponding to entries in y greater than the cutoff value are classified as false [true?], while those below the cutoff value are classified as true [false?]. These assigned class labels are compared with the true values in the classes slice and used to calculate the FPR and TPR.

roc.go
```diff
@@ -0,0 +1,136 @@
...
+// entries in y greater than the cutoff are classified as
+// false, and these assigned class labels are compared to
+// classes to calculate FPR and TPR.
+// weights being nil results in the unweighted case.
```
btracey: If weights is nil, all weights are treated as 1.

btracey: You need white space in between these lines. Godoc doesn't add a line break unless there is an empty line.

roc.go
```diff
@@ -0,0 +1,136 @@
...
+// n of zero results in all possible cutoffs being calculated.
```
btracey: I'm sorry if I missed the discussion, but is this really want we want? len(y) is so much clearer, and not very much work.

btracey: I see. The challenge is that there are len(y)+1 cutoffs.

stat.go
kortschak: Nah, I just like having history encapsulated in the language I speak.

roc.go
btracey: Zero seems fine in that case.

roc.go
```diff
@@ -0,0 +1,136 @@
...
+// More details about ROC curves are available at
+// https://en.wikipedia.org/wiki/Receiver_operating_characteristic
+func ROC(n int, y []float64, classes []bool, weights []float64) (tpr, fpr []float64) {
+	if len(y) != len(classes) {
+		panic("stat: slice length mismatch")
+	}
+	if !sort.IsSorted(sort.Float64Slice(y)) {
```
btracey: if !sort.Float64sAreSorted(y)

roc.go
```diff
@@ -0,0 +1,136 @@
...
+		panic("stat: input must be sorted")
+	}
+
+	var iw, tol float64
+	if n != 0 {
+		if n < 2 {
+			panic("stat: cannot less than 2 points on a ROC curve")
```
btracey: cannot use fewer than

roc.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+		iw = (y[len(y)-1] - y[0]) / float64(n-1)
+		if iw <= 0 {
+			panic("stat: y constant or not sorted")
+		}
+		tol = y[0] + iw
+	}
+
+	if n > (len(y) + 1) {
```
btracey: Sorry I'm being daft, but how does this work? If there are 6 values, and n = 4, this means that TPR and FPR are allocated to be of length 7. On the other hand, your tests pass...

btracey: Oh, I see, they're sliced at the bottom. I don't see a good reason to do that. We know how long the slices should end up.

roc.go
```diff
@@ -0,0 +1,136 @@
...
+// For a given cutoff value, observations corresponding to
+// entries in y greater than the cutoff are classified as
+// false, and these assigned class labels are compared to
+// classes to calculate FPR and TPR.
+// weights being nil results in the unweighted case.
+// n of zero results in all possible cutoffs being calculated.
+// More details about ROC curves are available at
+// https://en.wikipedia.org/wiki/Receiver_operating_characteristic
+func ROC(n int, y []float64, classes []bool, weights []float64) (tpr, fpr []float64) {
+	if len(y) != len(classes) {
```
btracey: Also need a check if weights != nil and len(weights) != len(y)

roc.go
btracey: Or rather, we know the maximum length slices can end up.

Is duplicating points in the curve worse than having an unpredictable return rate? I would guess not.

roc.go
```diff
@@ -0,0 +1,136 @@
...
+		tpr = make([]float64, n)
+		fpr = make([]float64, n)
+	} else {
+		tpr = make([]float64, len(y)+1)
+		fpr = make([]float64, len(y)+1)
+	}
+	newEntry := true
+	var j int
+	var nPositive, nNegative float64
+	for i, u := range classes {
```
btracey: This algorithm is very confusing to me.

Looking at the first if statement, it appears that newEntry is changed every other iteration. This doesn't make any sense. Upon a deeper look (after saying 'really? how does this possibly work', downloading your code and implementing a few more test cases), I see that this isn't really what's happening.

Point being, there must be a cleaner algorithm to do these implementations. For the moment, this would be helped by
- Giving 'j' a better name, like bin
- It feels like newEntry isn't actually necessary. Maybe you just need a "bin" and "prevBin". I find the name newEntry not very helpful, and given what I know the algorithm needs to do, it's not clear why such a boolean is necessary (which makes it more difficult)
- It's confusing that all of the data is assigned to by "j+1". I think it would be better to have j just start at 1 with a comment "// the initial bin is known to have 0 fpr and 0 tpr
- For the time being, it would be helpful to have the weights == nil check inside the inner to help see the algorithm better.
- Rather than check if n == 0, instead before the loop say that if n == 0, n = len(y)+1

roc.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+		}
+
+	}
+	if weights == nil {
+		nNegative = 1 / (float64(len(y)) - nPositive)
+	} else {
+		nNegative = 1 / nNegative
+	}
+	nPositive = 1 / nPositive
```
btracey: Just divide tpr and fpr instead of having the inverses.

stat.go
```diff
@@ -864,17 +864,67 @@ type weightSorter struct {
...
+// corresponding weights and boolean labels so that the x data are sorted.
+// The data is sorted in place.
```
btracey: Need newlines

stat.go
```diff
@@ -864,17 +864,67 @@ type weightSorter struct {
...
+// Weights and labels may be nil, if either is non-nil it
+// must have the same length as x.
+func SortWeightedLabelled(x []float64, labels []bool, weights []float64) {
+	if labels == nil {
```
btracey: instead of the else, return

if labels == nil {
    SortWeighted(x, weights)
    return
}
if weights == nil {
     ...
    return 
}
rest

stat_test.go
```diff
@@ -1264,6 +1265,65 @@ func TestSortWeighted(t *testing.T) {
 	}
 }
 
+func TestSortWeightedLabelled(t *testing.T) {
```
btracey: Spelling

btracey:
Sorry for the code review delay. I thought you had just implemented the sorting and not the other issues. Feel free to use an explicit "PTAL" in the future (please take another look).

We're getting there, thanks for your persistence.

roc.go
kortschak: This is my approach. Division is more costly that multiplication - in this case is probably doesn't make a great deal of difference, but if I'm dividing by the same value multiple times, I usually multiply by the reciprocal.

btracey: Okay. Can we call them invNegative and invPositive then?

roc.go
Armadilloa16: Yeah, I could swap them, which would more or less be equivalent to using !classes. Seems pretty equivalent really.

roc.go
Armadilloa16: Yep, also that calculating len(y)+1 equally spaced cutoffs is not the same as calculating the len(y)+1 unique (FPR,TPR) pairs. So thats why I made that the behavior when n==0, because essentially this is `special' -- in this case, the cutoffs are _not_ equally spaced in y, they are equally spaced in number of observations if that makes sense.

roc.go
Armadilloa16: I could improve this actually by adding the n == 0 as a special case. Is that worth doing?

Armadilloa16: But avoiding duplicate values is useful for three reasons: 
1. it makes calculating AUC slightly more trivial, 
2. it avoids ambiguity when duplicate values are from different classes, and
3. it makes implementing the n equally spaced cutoffs calculation the same as the 
all possible cutoffs calculation, which means we don't need to split into these two cases and have separate code for them.

stat.go
Armadilloa16: Also in SortWeighted then?

roc.go
Armadilloa16: I've made changes according to your dot points 1, 3, and 4.

Your second dot-point: newEntry became neccessary when we wanted to include arbitrary number of cutoffs and removal of duplicate values. Essentially it is because the first time we move to a new bin we need to copy the previous values, but if we havent just moved to a new bin then we just need to add to the current values and not copy from the previous values again. One possible solution would be to iterate i within the loop to skip the newEntry==true cases and do them manually when they are detected with 
`if y[i] != y[i+1]`
or
`if y[i+1] > tol . `
I'm not sure if this solution would be cleaner or not though, I like that the loop
`for i, u := range classes`
loop actually goes through len(y) loops and this would change that. 

Your last dot point, as I mentioned above n==0 is not the same as n=len(y)+1, 
n==len(y)+1 will produce FPR and TPR values for len(y)+1 equally spaced cutoffs
and this is not the same thing as producing FPR and TPR values for all possible
cutoffs (the behaviour when n==0). If the values in y are equally spaced, then these
two things are equivalent but otherwise they cannot be guaranteed to be the same, 
so I have deliberately left them as seperate cases.

Armadilloa16:
Thank you for all the feedback! and sorry for being so ignorant, like I think 
I mentioned awhile back I am relatively new to both golang and software development 
in general but I am learning alot so thank you all for your patience and comments! :)

roc.go
```diff
@@ -0,0 +1,145 @@
...
+//
+// For a given cutoff value, observations corresponding to
+// entries in y greater than the cutoff value are classified as
+// false, while those below (or equal to) the cutoff value are classified
+// as true. These assigned class labels are compared with the true values
+// in the classes slice and used to calculate the FPR and TPR.
+//
+// If weights is nil, all weights are treated as 1.
+//
+// n of zero results in all possible cutoffs being calculated.
```
btracey: There needs to be some discussion that the length of the returned slices may not be exactly n because of duplicate values and large weights.

roc.go
```diff
@@ -0,0 +1,145 @@
...
+			panic("stat: cannot calculate fewer than 2 points on a ROC curve")
+		}
+		incWidth = (y[len(y)-1] - y[0]) / float64(n-1)
+		if incWidth <= 0 {
+			panic("stat: input constant")
+		}
+		tol = y[0] + incWidth
+	}
+
+	if n > (len(y) + 1) {
```
btracey: If n < len(y), isn't it also an upper bound on the slice length? Can't this be 

```` 
if n == 0{
   tpr = make([]float64, len(y)+1)
}else{
   tpr = make([]float64, n)
}

File: roc.go [hunk 51]
```diff
@@ -0,0 +1,145 @@
...
+		tpr = make([]float64, n)
+		fpr = make([]float64, n)
+	} else {
+		tpr = make([]float64, len(y)+1)
+		fpr = make([]float64, len(y)+1)
+	}
+	newEntry := true
+	var bin int = 1 // the initial bin is known to have 0 fpr and 0 tpr
+	var nPositive, nNegative float64
+	for i, u := range classes {
```
btracey: I still think this can be better refactored. It has a long if statement that makes changes to newEntry which are non-permanent, because they are changed later.

It really feels like this can read something like
````
for i, u := range classes { 
     newBin = // check new bin
     if newBin() {
           bin++
           tpr[bin] = tpr[bin -1]
           fpr[bin] = fpr[bin-1]
     }
     newPos, newNeg := 1, 0
      if weights != nil {
           newPos = weights[i]
      }
      if !u {
          newPos, newNeg = newNeg, newPos
     }
     nPositive += newPos
     nNegative += newNeg
     tpr[bin] += newPos
    fpr[bin] += newNeg
}
````

File: roc.go (hunk 50 above)
Comment (Armadilloa16) [inline roc.go]: Yep, fixed that already actually then saw your comment haha -- sorry about that, should have realised that earlier.

File: roc.go (hunk 51 above)
Comment (Armadilloa16) [inline roc.go]: Yeah I've been thinking about it abit and I'm sure there is a better implementation but I hadn't thought it could be reduced this much -- good idea, I'm going to think about this now and see what I come up with.

**Comment by Armadilloa16** [comment]
Wow, ok that is just off the charts better -- thanks for the tips @btracey. I took your suggestion and modified it, got rid of newEntry entirely, but kept the ugly triple if at the end to deal with duplicates. Its so different from how it was I'm slightly skeptical, but its passing the tests... haha

**Comment by Armadilloa16** [comment]
Managed to combine the nested ifs so there are are only two nested levels now.

File: roc.go [hunk 52]
```diff
@@ -0,0 +1,110 @@
...
+	if len(y) != len(classes) {
+		panic("stat: slice length mismatch")
+	}
+	if weights != nil && len(y) != len(weights) {
+		panic("stat: slice length mismatch")
+	}
+	if !sort.Float64sAreSorted(y) {
+		panic("stat: input must be sorted")
+	}
+
```
btracey: This code panics if len(y) == 0, and has incWidth = 0 if len(y) == 1. I don't think we want to panic in this cases, so it seems like we need some special cases.

roc.go
```diff
@@ -0,0 +1,110 @@
...
+	var incWidth, tol float64
+	if n != 0 {
+		if n < 2 {
+			panic("stat: cannot calculate fewer than 2 points on a ROC curve")
+		}
+		incWidth = (y[len(y)-1] - y[0]) / float64(n-1)
+		if incWidth <= 0 {
```
btracey: I don't think this should panic. If incWidth == 0, then there's only one point in the middle of the ROC curve. May as well compute that point and return the set of three points.

roc.go
```diff
@@ -0,0 +1,110 @@
...
+		fpr = make([]float64, n)
+	} else {
+		tpr = make([]float64, len(y)+1)
+		fpr = make([]float64, len(y)+1)
+	}
+	var bin int = 1 // the initial bin is known to have 0 fpr and 0 tpr
+	var nPos, nNeg, posWeight, negWeight float64
+	for i, u := range classes {
+
+		posWeight, negWeight = 0, 1
```
btracey: Create posWeight and negWeight here, not up above.

roc.go
```diff
@@ -0,0 +1,110 @@
...
+			negWeight = weights[i]
+		}
+		if u {
+			posWeight, negWeight = negWeight, posWeight
+		}
+		nPos += posWeight
+		nNeg += negWeight
+		tpr[bin] += posWeight
+		fpr[bin] += negWeight
+
```
btracey: A more elaborate comment would be clearer up here than the half comments in the if.

```` Assess if the bin needs to be updated. If n is zero, the bin is always updated unless the consecutive y values are equal. Otherwise, the bin must be evaluated if the y value is sufficiently large````

File: roc.go [hunk 56]
```diff
@@ -0,0 +1,110 @@
...
+			// Deal with empty bins
+			for i != (len(y)-1) && y[i+1] > tol {
+				tol += incWidth
+				bin++
+				tpr[bin] = tpr[bin-1]
+				fpr[bin] = fpr[bin-1]
+			}
+		}
+	}
+	if n == 0 {
```
btracey: I'm pretty sure tpr and fpr has to be sliced in all cases, not just n == 0. It can be that bins are skipped, for example if the gap between one y value and the next is greater than  tol.

roc.go
Armadilloa16: Yeah I'll add special cases for these as well as tests for them. Thinking about it made me realise another thing -- if all the values in classes are equal (all true or all false), then the last entry of fpr or tpr won't be 1, it will be zero, and maybe we should mention that in the description as well?

roc.go
Armadilloa16: Yep I like that.

roc.go
Armadilloa16: Nope, my intention was that if n is specified and is not zero, then ROC should return slices of length n. So if there is a gap between one y value and the next that is greater than incWidth then that will result in consequtive entries in fpr and tpr that are equal -- so the same point on the ROC curve being repeated in the output. This makes sense to me because the concept of specifying n is that you are essentially specifying a list of n cutoffs to use, it just so happens that two (or more) of those cutoffs result in the same outcome. Does that make sense? The updated description of ROC should reflect that now.

roc.go
```diff
@@ -50,22 +50,36 @@ func ROC(n int, y []float64, classes []bool, weights []float64) (tpr, fpr []floa
...
-			panic("stat: input constant")
+		if incWidth == 0 {
+			for _, u := range classes {
+				if u {
+					tpr[n-1] = 1
+				} else {
+					fpr[n-1] = 1
+				}
+			}
+			return
```
Armadilloa16: There has to be a better way to do this. Essentially I want 
```
if any(classes) {
  tpr[n-1]=1
}
if any(!classes){
  fpr[n-1]=1
}
```
I just don't know how to do `any' in go

roc.go
```diff
@@ -60,8 +60,14 @@ func ROC(n int, y []float64, classes []bool, weights []float64) (tpr, fpr []floa
...
 					tpr[n-1] = 1
+					if fpr[n-1] == 1 {
+						break
+					}
 				} else {
 					fpr[n-1] = 1
+					if tpr[n-1] == 1 {
+						break
+					}
 				}
```
Armadilloa16: Honestly don't even know if this is better or worse. Its faster, I suppose.

Armadilloa16:
Ok, added the special cases, tests for them, and comments. PTAL @btracey 
There is one special case that I solved in a really ugly way that I hope there is an easier way to solve.
Alternatively we could just decide to do something different in that case. Its a pretty specific case that its addressing (when all the values in classes are equal -- either all false or all true).

roc.go
```diff
@@ -0,0 +1,132 @@
...
+			for i != (len(y)-1) && y[i+1] > tol {
+				tol += incWidth
+				bin++
+				tpr[bin] = tpr[bin-1]
+				fpr[bin] = fpr[bin-1]
+			}
+		}
+	}
+	if n == 0 {
+		tpr = tpr[0:(bin + 1)]
```
kortschak: Omit the 0 index in these slice operations.

roc.go
btracey: Ah, I see, I missed the "for" loop, and read it as an if. Add a comment that the for loop fills in the "skipped" bins. Wouldn't it make sense to do that for the n case as well, so that then the length is perfectly predictable?

I'm pretty sure your code is correct, but please add a "bottom" heavy test like
y = []float64{0.01, 0.02, 0.03, 0.04,0.05,0.06 10}, where n = 5

roc.go
```diff
@@ -0,0 +1,132 @@
...
+// assigned class labels are compared with the true values in the classes
+// slice and used to calculate the FPR and TPR.
+//
+// If weights is nil, all weights are treated as 1.
+//
+// n of zero results in all possible cutoffs being calculated -- this will
+// result in fpr and tpr having length len(y)+1 if y has no duplicated
+// values and one greater than the number of unique values in y in
+// general. n >= 2 will result in fpr and tpr having length n.
+// n of one or negative values of n are not supported.
```
btracey: "ROC will panic if n is equal to 1 or less than 0"

roc.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	var incWidth, tol float64
+	if n != 0 {
+		if n < 2 {
+			panic("stat: cannot calculate fewer than 2 points on a ROC curve")
+		}
+		tpr = make([]float64, n)
+		fpr = make([]float64, n)
+		if len(y) == 0 {
```
btracey: I think we should return before making the slices. I'd rather return them as nil than as a list of zeros. Please have an explicit "return nil, nil" -- it's becoming gonum policy to only have explicit returns.

roc.go
```diff
@@ -0,0 +1,132 @@
...
+					fpr[n-1] = 1
+					if tpr[n-1] == 1 {
+						break
+					}
+				}
+			}
+			return
+		}
+		tol = y[0] + incWidth
+	} else {
```
btracey: Could you add the shorter case first? I always find them easier to read.

roc.go
```diff
@@ -0,0 +1,132 @@
...
+				} else {
+					fpr[n-1] = 1
+					if tpr[n-1] == 1 {
+						break
+					}
+				}
+			}
+			return
+		}
+		tol = y[0] + incWidth
```
btracey: Could you move this to before the if statement? it's hard to see otherwise.

roc.go
```diff
@@ -0,0 +1,132 @@
...
+			panic("stat: cannot calculate fewer than 2 points on a ROC curve")
+		}
+		tpr = make([]float64, n)
+		fpr = make([]float64, n)
+		if len(y) == 0 {
+			return
+		}
+		incWidth = (y[len(y)-1] - y[0]) / float64(n-1)
+		if incWidth == 0 {
+			for _, u := range classes {
```
btracey: Why do we need to check if there's at least one positive and negative value? I would think we could just set the endpoints to (0,0), and (1,1) if they're all the same. I agree it's meaningless if there are no false data, but it's also meaningless if there are no true data, and this current algorithm has a bunch of (0,0) locations. I'm just not sure what looping through this data gets us.

Armadilloa16: Not much, just information about if they are all true or all false basically. The reason I decided to do it is because in the case that all the classes are true or all false but incWidth != 0, then ROC will return either tpr or fpr respectively as all zeros (including the last entry). So I was just trying to be consistent with that behaviour -- otherwise I agree its pretty meaningless. 

An alternative solution to being consistent is to force the last values to always be 1 -- and now I think about it I like that solution better, especially as we state in the description that the ROC curve will always end at (1,1).

Armadilloa16:
Done. I just set the final values of tpr and fpr to 1. Figure that makes as much sense as anything else in the special case classes is either all true or all false. Sound about right @btracey ?

roc.go
Armadilloa16: Umm no, so adding zero-width bins between duplicate y values is different to skipping equal-width bins that are empty because they are between y values. So unfortunately the length of the output in the n==0 case is not perfectly predictable, but in the n!=0 case the output will always be length n. 

I'll add a bottom heavy test case though, sure. I should probably also add some tests where the expected output is nil actually now I think about it.

roc.go
```diff
@@ -0,0 +1,123 @@
...
+			tpr[n-1] = 1
+			fpr[n-1] = 1
+			return
+		}
+	}
+
+	var bin int = 1 // the initial bin is known to have 0 fpr and 0 tpr
+	var nPos, nNeg float64
+	for i, u := range classes {
+
```
btracey: Remove this newline.

btracey:
LGTM with the newline removed. Do you have more comments @kortschak ?

Otherwise, the last thing is to fixup all of the commits to make it one commit instead of many. I find this link useful https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request.

Armadilloa16:
Yeah cool I figured out how to use 
`git rebase -i` 
awhile back now I just gotta remember.
I'll try it now for practice. If @kortschak 
has more comments I can do it again
afterwards. :)

roc.go
```diff
@@ -0,0 +1,122 @@
...
+//
+// For a given cutoff value, observations corresponding to entries in y
+// greater than the cutoff value are classified as false, while those
+// below (or equal to) the cutoff value are classified as true. These
+// assigned class labels are compared with the true values in the classes
+// slice and used to calculate the FPR and TPR.
+//
+// If weights is nil, all weights are treated as 1.
+//
+// n of zero results in all possible cutoffs being calculated -- this will
```
kortschak: Don't use `--` in comments. It has not markup meaning to godoc. I think this para could be rewritten to make it more English, since at the moment it reads more like code than prose.

Armadilloa16:
Not to muddy things any further... but are we happy with the `n' solution -- calculating for n equally spaced cutoffs (this is the solution some other packages use)? Now I think about it allowing for cutoffs to be specified, which would include this (as equally spaced cutoffs could be specified) is more general and might be better. Of course seeing as the current solution seems acceptable we could also leave it and then I could write an alternative and put in a pull request for that at some later point.

roc.go
```diff
@@ -26,11 +26,10 @@ import "sort"
...
-// n of zero results in all possible cutoffs being calculated -- this will
-// result in fpr and tpr having length len(y)+1 if y has no duplicated
-// values and one greater than the number of unique values in y in
-// general. n >= 2 will result in fpr and tpr having length n.
-// ROC will panic if n is equal to one or less than 0.
+// n of zero results in all possible cutoffs being calculated, resulting
+// in fpr and tpr having length one greater than the number of unique 
+// values in y. n greater than one will result in fpr and tpr having 
+// length n. ROC will panic if n is equal to one or less than 0.
 //
```
Armadilloa16: I tried re-writing the comments, what do you think @kortschak ?

roc.go
```diff
@@ -0,0 +1,121 @@
...
+//
+// For a given cutoff value, observations corresponding to entries in y
+// greater than the cutoff value are classified as false, while those
+// below (or equal to) the cutoff value are classified as true. These
+// assigned class labels are compared with the true values in the classes
+// slice and used to calculate the FPR and TPR.
+//
+// If weights is nil, all weights are treated as 1.
+//
+// n of zero results in all possible cutoffs being calculated, resulting
```
kortschak: When n is zero all possible cut-offs are calculated,...

roc.go
```diff
@@ -0,0 +1,121 @@
...
+// in fpr and tpr having length one greater than the number of unique
+// values in y. n greater than one will result in fpr and tpr having
```
kortschak: When n is greater than one, fpr and tpr will be returned with length n.

Armadilloa16:
Ok I've made the improvements you both suggested @btracey @kortschak and cleaned up the commit history. Also over the weekend I thought about it and realised that we can make the code easier to understand, simpler, and also more general by replacing the input n int with cutoffs []float64, specifying cutoffs, so I implemented that version and you can take a look at it here:
https://github.com/Armadilloa16/stat/commit/1a94226412eb3a7d63d0315c4067c8a0de062293
If you like it, I'll merge it into this pull request? cutoffs == nil produces the same behaviour as n==0, but this can also be done by selecting the correct cutoffs as well now.

Armadilloa16:
@kortschak @btracey Hey sorry I got distracted and really busy for a week or two there.

### ROC Function:
Where we left this, I think I addressed all your points, and then also suggested an alternative 
in which the thresholds can be explicitly given, as an alternative to the current version where just 
the number of thresholds is given and they are equally spaced. Or if the number is zero then they 
can be un-equally spaced and cover every possible case. Inputting the thresholds themselves results in slightly simpler code, and includes both these cases, but also allows for arbitrary unequally spaced thresholds to be specified. A link to a branch I made showing how this would look is above.

### Confusion Matrix:
In the meantime, I was thinking about it and realised that maybe would be better to simply calculate the entire confusion matrix (four outputs, fp, tp, fn, tn) of which this would be a subset -- i.e. fpr = fp / (fp + tn) and tpr = tp / (tp + fn). On the other hand this probably belongs in its own function, as its purpose would apply to different situations -- mostly ROC curves are calculated from samples which are not representative of the prevalence (tp + fn) / (tp + fn + tn + fp), and the ROC curve is being calculated because fpr and tpr are prevalence-independent. In these cases you would not want to calculate the confusion matrix directly from your data, but rather to infer it from your fpr, tpr values and a known prevalence value from some cohort study or something. So thats probably the better way to go about that. Or maybe two different functions, one for calculating the confusion matrix from data, one to infer it from a ROC curve using a known prevalence value.

What do you think?

btracey:
LGTM

Let's get this in as is since it's already been a lot. We can discuss changes and additions elsewhere.

kortschak:
There's an accidental file inclusion that needs to be removed, and the commits need to be squashed and rebased on to master. Then LGTM.

kortschak:
Please `git rm  .roc.go.swp ` and `git commit --amend` and then rebase against gonum/stat master with a forced push to your fork's branch.

Armadilloa16:
Done. PTAL @kortschak

kortschak:
LGTM

@btracey Do you want to merge? otherwise I'll do it in the morning.

PR #99, opened 2016-04-20
Addmargsing

btracey:


btracey:
This looks like it duplicates functionality with MarginalNormal, but the single element version doesn't require setSigma to be called, and generates less garbage. This change alone lead to a 2-4x speedup on my personal code.

kortschak:
Since this replicates functionality, include a comparison of the two in the tests.

btracey:
Done. 
````
BenchmarkMarginalNormal10-8         	 1000000	      1341 ns/op
BenchmarkMarginalNormalReset10-8    	  300000	      4992 ns/op
BenchmarkMarginalNormalSingle10-8   	50000000	        30.2 ns/op
````

The algorithm is a different complexity order, so the gap will only grow larger with larger N.

kortschak:
Sorry, I meant a comparison of the results of the two in the case where they are equivalent, not benchmark.

btracey:
Done.

kortschak:
LGTM

PR #102, opened 2016-05-19
stat: add AUC

sbinet:


btracey:
Does this really belong here? Shouldn't this be in quad somehow?

kortschak:
From a usability perspective, it probably does belong here. People looking for AUROC will have found ROC here, looking in quad may be surprising requirement. The alternative is to put it there (if that's where it belongs) and have an `ExampleROC_AUROC`.

btracey:
I was definitely thinking an example would be in stat, along with possibly an explicit pointer in the ROC documentation. It seems like it's quad's job to handle 1-d function integration, in particular, one may want to estimate the integral in different ways (i.e. spline fit), and we wouldn't want all of those alternatives implemented in stat.

kortschak:
SGTM

sbinet:
done. needs https://github.com/gonum/quad/pull/17

kortschak:
The commit message and PR title need to be changed to 'stat: add AUC example' and the example name does not need the qualifier. The actual implementation of AUC here should be deleted.

sbinet:
sure. done. (I was actually waiting for gonum/quad#17 to be reviewed before going back on this one :)

sbinet:
PTAL.

I renamed the example `ExampleROC_auc` so it would appear as `ROC (auc)` in `godoc`.
I could name it `ExampleROC_aUC` so it would appear as `ROC (AUC)`, though...

kortschak:
I think ExampleROC_aUC (does it really have to be lc in the first pos?).

Note also that in #110 ROC changes.

sbinet:
yep, it needs to be lower case.
(_wrt_ #110, I'll adjust if it goes in first)

sbinet:
`ExampleROC_aUC` it is.

sbinet:
PTAL (I've rebased and squashed)

roc_example_test.go
```diff
@@ -36,3 +37,19 @@ func ExampleROC_weighted() {
 	// true  positive rate: [0 0.5 0.5 1 1 1 1]
 	// false positive rate: [0 0 0.25 0.25 0.5 0.75 1]
 }
+
+func ExampleROC_aUC() {
+	y := []float64{0.1, 0.35, 0.4, 0.8}
+	classes := []bool{true, false, true, false}
+
+	tpr, fpr := stat.ROC(0, y, classes, nil)
+	auc := integrate.Trapezoidal(fpr, tpr)
```
btracey: Are people going to know this means "Area Under Curve"? I wouldn't have before this PR. Maybe add a comment above the call to Trapezoid?

sbinet: done

btracey:
LGTM

PR #103, opened 2016-05-23
AddTransformNormal

btracey:


btracey:
PTAL @sbinet @kortschak 

Calling this TransformNormal because I think there is room for a Transform() that works on slices generated from the standard uniform.

This is useful for comparing results under two different Gaussians. I can generate a bunch of vectors from a standard normal, and then test the results using Transform with different Gaussians.

distmv/normal.go
```diff
@@ -354,3 +355,31 @@ func (n *Normal) setSigma() {
...
+	dst = reuseAs(dst, n.dim)
+	if len(dst) != len(src) {
+		panic(badInputLength)
+	}
+	n.transformNormal(dst, src)
+	return dst
+}
+
+// transformNormal performs the same operation as TransformNormal except no
+// safety checks are performed and both must be non-nil.
```
sbinet:
```
-and both must be non-nil.
+and both dst and src must be non-nil.
```
?
(or "both input slices must be non-nil" ?)

sbinet:
LGTM (but I am not an expert, by far, so... please wait for another LGTM)

distmv/normal.go
btracey: Done

btracey:
Ping @vladimir-ch @jonlawlor

jonlawlor:
LGTM.

distmv/normal.go
```diff
@@ -354,3 +355,31 @@ func (n *Normal) setSigma() {
 		n.sigma.FromCholesky(&n.chol)
 	})
 }
+
+// TransformNormal transforms the vector src, generated from a multidimensional
+// standard normal, to one generated under the distribution specified by the
+// receiver. If dst is non-nil, the result will be stored into dst, otherwise
+// a new slice will be allocated. TransformNormal will panic if the length of
+// src is not the dimension of the Normal, or if dst is non-nil and len(src) !=
```
vladimir-ch: s/of the Normal/of the receiver/  ?
I would write the comparison reversed `len(dst) != len(src)`.

distmv/normal.go
```diff
@@ -354,3 +355,31 @@ func (n *Normal) setSigma() {
...
+// len(dst).
+func (n *Normal) TransformNormal(dst, src []float64) []float64 {
```
vladimir-ch: For my understanding: should or could this be a standalone function?

distmv/normal.go
```diff
@@ -354,3 +355,31 @@ func (n *Normal) setSigma() {
...
+	if len(src) != n.Dim() {
```
vladimir-ch: Use either `n.Dim()` or `n.dim` everywhere?

distmv/normal.go
btracey: I could be, but I don't think it should be. In my mind, functions should use the public interface of the types, that is, they could be written in this package or somewhere else. Methods, on the other hand, in my mind make use of the private data to perform the operation. Under that logic, this should be a method.

btracey: I guess in this case it's unclear. You could write this method outside, if you took your own Cholesky decomposition. It's a lot more efficient to just use the one we already have. Not sure which way that sways in this case. @kortschak ?

kortschak: If it's on the `*Normal` it doesn't need to be called `TransformNormal`, it can just be `Transform`.

The normal logic for determining whether something is a package function or method is mainly one of two things: 1) namespacing and 2) making generic code. The first is obvious, the second is when you want to use specific implementations to fit within a generic framework provided by package-level functions or type methods.

Here, it looks like the `Transform` is the generic interface and `*Normal` is the specific implementation. So, with the name alteration, I'd make this a method.

jonlawlor: The doc suggests that other distributions would have a similar `Transform(Normal)?` method which takes a multidimensional standard normal and returns values from the receiver's distribution, so you might need `TransformUniform` as well.  In that case the name repetition is just a coincidence and is appropriate.

I might be reading too much into the docs though.

btracey: I think the base would be Transform(dst, src) where src is drawn from a uniform distribution

kortschak: That's not emphasised by the docs. I think it could be clearer. In that case, still method, but with the same name or TransformNorm, and the uniform source case TransformUnif to be explicit.

btracey: In order to emphasize it more, do you just want it not inside a paranthetical expression? The fact that it's generated from a mvn is in the first line of the doc.

kortschak: Yes, it does feel parenthetic. This is probably because of the repetition.

kortschak: A possibility is to rename the src parameter to norm.

distmv/normal.go

distmv/normal.go
btracey: Done, and comment updated.

distmv/normal.go

distmv/normal.go
```diff
@@ -354,3 +355,32 @@ func (n *Normal) setSigma() {
 		n.sigma.FromCholesky(&n.chol)
 	})
 }
+
+// TransformNormal transforms the vector norm that has been generated from a
```
kortschak: TransformNorm

distmv/normal.go
```diff
@@ -354,3 +355,32 @@ func (n *Normal) setSigma() {
...
+// standard multidimensional normal into a vector that has been generated under
+// the distribution of the receiver.
+//
+// If dst is non-nil, the result will be stored into dst, otherwise a new slice
+// will be allocated. TransformNormal will panic if the length of norm is not
```
kortschak: Here too.

distmv/normal.go
```diff
@@ -354,3 +355,32 @@ func (n *Normal) setSigma() {
...
+	}
+	dst = reuseAs(dst, n.dim)
+	if len(dst) != len(norm) {
+		panic(badInputLength)
+	}
+	n.transformNormal(dst, norm)
+	return dst
+}
+
+// transformNormal performs the same operation as TransformNormal except no
```
kortschak: Probably also here and in the signature (and so the call above).

distmv/normal.go
btracey: Done

distmv/normal.go
btracey: Done

distmv/normal.go

btracey:
PTAL

vladimir-ch:
Sorry for not paying attention earlier. Is `TransformNorm` a better name than `TransformNormal`? I don't like that the abbreviation becomes a completely different concept (but I can bend easilly).

kortschak:
I'm not particularly fussed. The reason I suggested `TransformNorm` over `TransformNormal` was that the parameter is called `norm` and norm is the usual name for normal operations in places like R.

What is the completely different concept that it becomes?

vladimir-ch:
Probably because of a different background, for me these two words don't mesh together. When I see `Norm`, I think of a norm of a vector. But then `Transform` doesn't fit, and I become confused. If you think it's fine and if it's used in other places (I know nothing about R), then I'm happy.

kortschak:
OK. MATLAB uses norm in the sense you expect. Maybe we should expand to `TransformNormal` and the parameter to `normal` to avoid any confusion. @btracey can make the call.

btracey:
Changed back to TransformNormal

distmv/normal.go
```diff
@@ -356,28 +356,28 @@ func (n *Normal) setSigma() {
 	})
 }
 
-// TransformNorm transforms the vector, norm, generated from a standard
+// TransformNormal transforms the vector, norm, generated from a standard
```
kortschak: s/norm/normal/

distmv/normal.go
```diff
@@ -356,28 +356,28 @@ func (n *Normal) setSigma() {
...
-	n.transformNorm(dst, norm)
+	n.transformNormal(dst, normal)
 	return dst
 }
 
-// transformNorm performs the same operation as TransformNorm except no
+// transformNormal performs the same operation as TransformNormal except no
 // safety checks are performed and both input slices must be non-nil.
-func (n *Normal) transformNorm(dst, norm []float64) []float64 {
+func (n *Normal) transformNormal(dst, norm []float64) []float64 {
```
kortschak: s/norm/normal/g

distmv/normal.go
btracey: Ah, shoot

btracey: Done

distmv/normal.go
btracey: Done

btracey:
ping

kortschak:
LGTM

PR #106, opened 2016-06-09
Rename to Score function and increase test coverage

btracey:


btracey:
Apparently this thing has a name. 

I chose Score and ScoreInput because usually the score function is with respect to the parameters, but in a few cases I've seen it talked about with respect to the input. Since the parameter version is much more common (see the wikipedia entry) it seems like it should get the base name. 

PTAL @kortschak

kortschak:
Link to WP article (at least for PR comments)? You mean [this](https://en.wikipedia.org/wiki/Score_%28statistics%29)?

I take it the colour and movement is for alphabetical ordering.

LGTM

btracey:
Yes, the color is from alphabetical.

PR #107, opened 2016-06-16
Add Bernoulli distribution

btracey:


distuv/bernouilli.go
```diff
@@ -0,0 +1,111 @@
+package distuv
+
+import (
+	"math"
+	"math/rand"
+)
+
+// Bernoulli represents a random variable whose value is 1 with probability p and
+// value of zero with probability 1-P. The value of P must be between 0 and 1.
+type Bernouilli struct {
```
vladimir-ch: Bernoulli

btracey: Ah shoot. I even tried to double check it....

btracey:
PTAL

distuv/bernouilli.go
```diff
@@ -0,0 +1,111 @@
+package distuv
```
kortschak: Header

distuv/bernouilli.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"math"
+	"math/rand"
+)
+
+// Bernoulli represents a random variable whose value is 1 with probability p and
+// value of zero with probability 1-P. The value of P must be between 0 and 1.
+type Bernoulli struct {
```
sbinet: for a few types in `gonum/*` we had a link to a wikipedia description of the mathematical concept.
should we follow the trend and add https://en.wikipedia.org/wiki/Bernoulli_distribution here as well?

sbinet:
as Bernoulli is Bernoulli and not Bernouilli, perhaps also rename the file `bernoulli.go` as well ? :)

sbinet:
perhaps add a test as for `lognormal` and `exponential` ?

btracey:
What do you mean by adding a test for those two?

sbinet:
not for them but like for these 2 distributions, add a test for Bernoulli :)

btracey:
Oh, "as for" those two

distuv/bernouilli.go
btracey: Done.

distuv/bernouilli.go
btracey: Done.

btracey:
Completely refactored testFullDist to make it work with discrete distributions. Also completely changed the PDF test because the old one was a load of trash.

distuv/distribution_test.go
```diff
@@ -9,135 +9,214 @@ import (
 	"sort"
 	"testing"
 
+	"github.com/btracey/quad"
```
sbinet: this should probably gonum's quad instead of yours.

btracey: Good catch. Done.

I can't win with goimports and quad. I deleted btracey/quad from existence, then it wanted github.com/gonum/quadold (which I had kept because of other branches). I changed the package name in quadold, and then it wanted gonum.org/pkg/quad

distuv/bernoulli.go
```diff
@@ -0,0 +1,132 @@
...
+
+// Prob computes the value of the probability distribution at x.
+func (b Bernoulli) Prob(x float64) float64 {
+	return math.Exp(b.LogProb(x))
+}
+
+// Quantile returns the inverse of the cumulative probability distribution.
+func (b Bernoulli) Quantile(p float64) float64 {
+	if p < 0 || p > 1 {
+		panic("dist: percentile out of bounds")
```
sbinet:
```
-dist
+distuv
```

btracey: Can I fix that in the next commit? It's a problem with all of the distributions.

sbinet: sure.

sbinet:
LGTM

distuv/distribution_test.go
```diff
@@ -10,134 +10,213 @@ import (
...
-	if !floats.EqualWithinAbsOrRel(median, f.Median(), tol, tol) {
-		t.Errorf("Median mismatch case %v: want: %v, got: %v", i, median, f.Median())
+type probLogprober interface {
+	Prob(x float64) float64
+	LogProb(x float64) float64
+}
+
+func checkMean(t *testing.T, i int, x []float64, m meaner, tol float64) {
+	mean := m.Mean()
+	if !floats.EqualWithinAbsOrRel(mean, m.Mean(), tol, tol) {
```
vladimir-ch: Late to the party, sorry. This effectively compares m.Mean() to m.Mean(), and x is not used.

vladimir-ch: I see, the line above should probably be `mean := stat.Mean(x, nil)`.

distuv/bernoulli.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// Prob computes the value of the probability distribution at x.
+func (b Bernoulli) Prob(x float64) float64 {
+	return math.Exp(b.LogProb(x))
+}
+
+// Quantile returns the inverse of the cumulative probability distribution.
+func (b Bernoulli) Quantile(p float64) float64 {
+	if p < 0 || p > 1 {
```
vladimir-ch: This would be nicer as `if p < 0 || 1 < p {`. Could you add it to #108 ?

distuv/bernoulli.go
```diff
@@ -0,0 +1,132 @@
...
+func (b Bernoulli) Mean() float64 {
+	return b.P
+}
+
+// Median returns the median of the probability distribution.
+func (b Bernoulli) Median() float64 {
+	p := b.P
+	q := 1 - p
+	switch {
+	case p < q:
```
vladimir-ch: I think that `q` is not necessary. These comparisons are the same as comparing `p` to 0.5. Could you add it to #108 ?

btracey: Done.

distuv/bernoulli.go
btracey: Done.

distuv/distribution_test.go
btracey: Yep. Done. Sorry.

PR #108, opened 2016-06-16
Fix error messages and improve tests.

btracey:
In the move from dist to distuv, many error messages were not fixed. This fixes the error messages to agree on the distuv: prefix and has them use the same panic string. Additionally, this improves the testing of continuous distributions to make sure PDF agrees with Quantile and Rand

### DIFF
--- a/distuv/distribution_test.go
+++ b/distuv/distribution_test.go
@@ -65,8 +65,13 @@ type probLogprober interface {
 	LogProb(x float64) float64
 }
 
+type cumulantProber interface {
+	cumulanter
+	probLogprober
+}
+
 func checkMean(t *testing.T, i int, x []float64, m meaner, tol float64) {
-	mean := m.Mean()
+	mean := stat.Mean(x, nil)
 	if !floats.EqualWithinAbsOrRel(mean, m.Mean(), tol, tol) {
 		t.Errorf("Mean mismatch case %v: want: %v, got: %v", i, mean, m.Mean())
 	}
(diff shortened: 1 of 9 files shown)


btracey:
PTAL

distuv/distribution_test.go
```diff
@@ -160,7 +168,51 @@ func checkPDFContinuous(t *testing.T, i int, x []float64, p probLogprober) {
 	}
 }
 
-func checkPDFDiscrete(t *testing.T, i int, xs []float64, p probLogprober, tol float64) {
+// checkProbCdfContinuous checks that the Prob, Rand, and Quantile are all consistent.
```
vladimir-ch: checkProbQuantContinuous

vladimir-ch: Ah, no, just reorder the comment into the usual go order.

vladimir-ch:
LGTM

distuv/distribution_test.go
btracey: Huh? I fixed the comment function name

vladimir-ch: Bear with me today. I had difficulties parsing the comment and I thought that there was another function called `checkProbCdfContinuous`. I parsed it as "A checks that B. C only checks that D. However, ... therefore _this_ checks also E." Also, what does `Rand` refer to here?

btracey: The xs are generated from Rand()

vladimir-ch: Wouldn't it be clearer if the samples were generated anew inside each check*() function?

sbinet:
LGTM

distuv/distribution_test.go
btracey: Probably, but generating the random numbers is slow, so the test time would increase significantly.

Issue #104, opened 2016-06-01
stat: ROC signature


Armadilloa16:
I came up with what I think is an improvement to the ROC function, 
in which the cutoffs are defined explicitly instead of the number of cutoffs n.
This allows for unequally-spaced cutoffs, allows for all cutoffs to be used
more naturally, and simplifies the code itself.

Is this a good idea -- should I put a PR in for it, @sbinet @kortschak ?

Also, I was thinking maybe ROC should output an error as well, instead of all
the panic statements? Or is it appropriate that it panic?

What I did is here, in case you wanted to look at the specifics of what I am 
suggesting:
https://github.com/Armadilloa16/stat/tree/roc

kortschak:
The panics are appropriate under the model we use; they are all things that can easily be checked prior to the call and are programmer errors if the invariants being checked are not upheld.

As far as the change goes, I'm not sure. If it went in it would make the common case (equal spacing) harder to use unless a helper that generates equal spacing were provided (potentially requiring an allocation per call). I'll leave this up to others to decide.

sbinet:
thinking out loud, wrt handling equal spacing better: why about having a config struct?

```go
type ROCConfig struct {
  CutOffs []float64
  N int
}
```

the zero value of `ROCConfig` would enable the old code path (`n==0`).

(but remember that I am not an expert on this matter so I don't really know what are the most common cases... and thus what to prioritize as an API)

kortschak:
I'd prefer two functions over a config struct.

Armadilloa16:
I like the idea of a config struct in principle, but it seems like overkill for the ROC function.
In this case I prefer either two functions, or just the one cutoffs function with a helper that generates equally spaced cutoffs (easy enough, just start at the minimum value minus eps and end at the maximum value, with equally spaced intervals between, for n >= 2).

sbinet:
could you update the PR with the helper function-to-generate-cutoffs option to see how that looks like?

thanks.

kortschak:
We might have troubles getting hold of Lyron for the next while. He has just submitted this thesis and will be away for a while.

Armadilloa16:
Hey yeah I'm heading off for awhile and I'll be intermittently out of contact until midway through August -- but when I'm back I'll definitely have a go at this, and some other things I have in the pipeline as well.

PR #110, opened 2016-06-29
ROC: Alternate Signature

Armadilloa16:
Found some free time, so I had a go at this. 

As per #104 the new ROC function takes cutoffs as an argument instead of the number of cutoffs n, and there is a helper function EquallySpaced for generating equallly spaced cutoffs. The need for a helper function to maintain the old functionality is abit clunky but this allows for particular cutoffs of interest to be defined which wasn't the case before, and I think the code is (slightly) easier to read.

PTAL @kortschak @sbinet :)

### DIFF
--- a/roc.go
+++ b/roc.go
@@ -20,7 +20,7 @@ import (
 //
 // For a given cutoff value, observations corresponding to entries in y
 // greater than the cutoff value are classified as false, while those
-// less than (or equal to) the cutoff value are classified as true. These
+// less than or equal to the cutoff value are classified as true. These
 // assigned class labels are compared with the true values in the classes
 // slice and used to calculate the FPR and TPR.
 //
@@ -29,8 +29,8 @@ import (
 // If cutoffs is nil or empty all possible cutoffs are calculated,
 // resulting in fpr and tpr having length one greater than the number of
 // unique values in y. Otherwise fpr and tpr will be returned with the
-// same length as cutoffs. EquallySpaced can be used to generate
-// equally spaced cutoffs.
+// same length as cutoffs. floats.Span can be used to generate equally 
+// spaced cutoffs.
 //
 // More details about ROC curves are available at
 // https://en.wikipedia.org/wiki/Receiver_operating_characteristic
@@ -42,17 +42,21 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr []fl
 		panic("stat: slice length mismatch")
 	}
 	if !sort.Float64sAreSorted(y) {
-		panic("stat: input must be sorted")
+		panic("stat: input must be sorted ascending")
 	}
 	if !sort.Float64sAreSorted(cutoffs) {
-		panic("stat: input must be sorted")
+		panic("stat: cutoff values must be sorted ascending")
 	}
 	if len(y) == 0 {
 		return nil, nil
 	}
 	var bin int
 	if len(cutoffs) == 0 {
-		cutoffs = make([]float64, len(y)+1)
+		if cutoffs == nil || cap(cutoffs) < len(y)+1 {
+		    cutoffs = make([]float64, len(y)+1)
+		} else {
+		    cutoffs = cutoffs[:len(y)+1]
+		}
 		cutoffs[0] = math.Nextafter(y[0], y[0]-1)
 		// Choose all possible cutoffs but remove duplicate values
 		// in y.
@@ -62,7 +66,7 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr []fl
 			}
 			cutoffs[bin+1] = u
 		}
-		cutoffs = cutoffs[0:(bin + 2)]
+		cutoffs = cutoffs[0:bin + 2]
 	}
 
 	tpr = make([]float64, len(cutoffs))
@@ -71,9 +75,9 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr []fl
 	var nPos, nNeg float64
 	for i, u := range classes {
 		// Update the bin until it matches the next y value
-		// (skip empty bins).
-		for (bin < len(cutoffs)) && (y[i] > cutoffs[bin]) {
-			if bin == (len(cutoffs) - 1) {
+		// skipping empty bins.
+		for bin < len(cutoffs) && y[i] > cutoffs[bin] {
+			if bin == len(cutoffs) - 1 {
 				break
 			}
 			bin++
@@ -105,23 +109,3 @@ func ROC(cutoffs, y []float64, classes []bool, weights []float64) (tpr, fpr []fl
 	return tpr, fpr
 }
 
-// EquallySpaced returns n equally spaced values,
-// beggining eps less than min and ending at max. n must be
-// at least two, and min must not be greater than max.
-func EquallySpaced(min, max float64, n int) (cutoffs []float64) {
-	if max < min {
-		panic("stat: max < min")
-	}
-	if n < 2 {
-		panic("stat: n too small")
-	}
-
-	cutoffs = make([]float64, n)
-	w := (max - min) / float64(n-1)
-	for i := range cutoffs {
-		cutoffs[i] = min + w*float64(i)
-	}
-	cutoffs[0] = math.Nextafter(min, min-1)
-
-	return cutoffs
-}
--- a/roc.go
+++ b/roc.go
@@ -29,7 +29,7 @@ import (
 // If cutoffs is nil or empty all possible cutoffs are calculated,
 // resulting in fpr and tpr having length one greater than the number of
 // unique values in y. Otherwise fpr and tpr will be returned with the
-// same length as cutoffs. floats.Span can be used to generate equally 
+// same length as cutoffs. floats.Span can be used to generate equally
 // spaced cutoffs.
 //
 // More details about ROC curves are available at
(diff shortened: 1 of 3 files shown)


roc.go
```diff
@@ -81,41 +89,39 @@ func ROC(n int, y []float64, classes []bool, weights []float64) (tpr, fpr []floa
...
 		tpr[i] *= invPos
 		fpr[i] *= invNeg
 	}
-	tpr[len(tpr)-1] = 1
-	fpr[len(fpr)-1] = 1
 
 	return tpr, fpr
 }
+
+// EquallySpaced returns n equally spaced values,
```
btracey: Users can just use floats.Span

roc.go
```diff
@@ -4,36 +4,37 @@
...
-// The input y must be sorted, and SortWeightedLabeled can be used in
-// order to sort y together with classes and weights.
+// The input y and cutoffs must be sorted, and values in y must correspond
+// to values in classes and weights. SortWeightedLabeled can be used to
+// sort y together with classes and weights.
 //
 // For a given cutoff value, observations corresponding to entries in y
 // greater than the cutoff value are classified as false, while those
-// below (or equal to) the cutoff value are classified as true. These
+// less than (or equal to) the cutoff value are classified as true. These
```
kortschak: Remove parens.

roc.go
kortschak: That's lovely. The code is almost identical, though there is a difference in the constraints on ordering and the value of lowest returned element; @Armadilloa16 can you explain why you need to make `cutoffs[0]` one ulp smaller?

If `floats.Span` is used `ROC` needs to check cutoff ordering is ascending or `float.Span` needs to prohibit descending spans (I don't think that is a good idea).

roc.go
```diff
@@ -43,35 +44,42 @@ func ROC(n int, y []float64, classes []bool, weights []float64) (tpr, fpr []floa
...
-		tpr = make([]float64, n)
-		fpr = make([]float64, n)
-		incWidth = (y[len(y)-1] - y[0]) / float64(n-1)
-		tol = y[0] + incWidth
-		if incWidth == 0 {
-			tpr[n-1] = 1
-			fpr[n-1] = 1
-			return
+	if !sort.Float64sAreSorted(cutoffs) {
+		panic("stat: input must be sorted")
```
kortschak: `panic("stat: cutoff values must be sorted ascending")`

roc.go
```diff
@@ -43,35 +44,42 @@ func ROC(n int, y []float64, classes []bool, weights []float64) (tpr, fpr []floa
 	if !sort.Float64sAreSorted(y) {
 		panic("stat: input must be sorted")
```
kortschak: `panic("stat: input must be sorted ascending")`

roc.go
```diff
@@ -43,35 +44,42 @@ func ROC(n int, y []float64, classes []bool, weights []float64) (tpr, fpr []floa
...
-			fpr[n-1] = 1
-			return
+	if !sort.Float64sAreSorted(cutoffs) {
+		panic("stat: input must be sorted")
+	}
+	if len(y) == 0 {
+		return nil, nil
+	}
+	var bin int
+	if len(cutoffs) == 0 {
```
kortschak: The condition here wrt the make call should be whether `cutoffs == nil` or `cap(cutoffs) < len(y)+1`. Otherwise just resize the zero-length slice to the correct size.

```
if cutoffs == nil || cap(cutoffs) < len(y)+1 {
    cutoffs = make([]float64, len(y)+1)
} else {
    cutoffs = cutoffs[:len(y)+1)]
}
```

roc.go
```diff
@@ -43,35 +44,42 @@ func ROC(n int, y []float64, classes []bool, weights []float64) (tpr, fpr []floa
...
 
-	var bin int = 1 // the initial bin is known to have 0 fpr and 0 tpr
+	tpr = make([]float64, len(cutoffs))
+	fpr = make([]float64, len(cutoffs))
+	bin = 0
 	var nPos, nNeg float64
 	for i, u := range classes {
+		// Update the bin until it matches the next y value
+		// (skip empty bins).
+		for (bin < len(cutoffs)) && (y[i] > cutoffs[bin]) {
```
kortschak: Remove the parens around the conditions.

roc.go
```diff
@@ -43,35 +44,42 @@ func ROC(n int, y []float64, classes []bool, weights []float64) (tpr, fpr []floa
...
 	}
 
-	var bin int = 1 // the initial bin is known to have 0 fpr and 0 tpr
+	tpr = make([]float64, len(cutoffs))
+	fpr = make([]float64, len(cutoffs))
+	bin = 0
 	var nPos, nNeg float64
 	for i, u := range classes {
+		// Update the bin until it matches the next y value
+		// (skip empty bins).
```
kortschak: s/(skip empty bins)/skipping empty bins./

roc.go
Armadilloa16: Yep, I'll just use Span instead. I knew there had to be a function somewhere that did that but I just didn't know where it was, thanks! As for why cutoffs[0] has to be slightly less than y[0] -- this is so that the ROC curve will start at (0,0) -- as in the description, y values less than or equal to the cutoff are classified as true. If the equality was in the other direction, you would need the last cutoff to be slightly greater than the maximum in order for the ROC curve to end at (1,1).

roc.go
Armadilloa16: cutoffs can be longer than len(y)+1 though -- this will simply result in repeated values on the ROC curve. What this should be checking for is if cutoffs is nil or empty, which is what len(cutoffs)==0 achieves -- given that this is the case though you are right I should check the cap in case calling make can be avoided.

Armadilloa16: Oh wait nvm I misunderstood what you where saying -- got it now.

roc.go
Armadilloa16: parens, cool.

if we wanted to force the ROC curve to always start at (0,0) and always end at (1,1) then your suggestion would work -- but this implementation allows for arbitrary cutoffs to be defined and so that is no longer necessarily the case. You could even pass a single cutoff to get a single fpr, tpr pair. Now I'm thinking about it there are alot of extra special cases I should add tests for...

kortschak: `for bin < len(cutoffs) && ((bin == 0 && y[i] >= cutoffs[0]) || y[i] > cutoffs[bin]) {`

roc_example_test.go
```diff
@@ -77,7 +79,8 @@ func ExampleROC_equallySpacedCutoffs() {
 	n := 9
 
 	stat.SortWeightedLabeled(y, classes, weights)
-	cutoffs := stat.EquallySpaced(y[0], y[len(y)-1], n)
+	cutoffs := make([]float64, n)
+	floats.Span(cutoffs, math.Nextafter(y[0], y[0]-1), y[len(y)-1])
```
kortschak: The use of Nextafter is a code smell. We need to get a logic that doesn't need it. See my last comment.

Armadilloa16: you could use 'floats.Span(cutoffs,y[0],y[len(y)-1])' and then just attach (0,0) to the beggining of the output from ROC.

Armadilloa16: In principle, if you wanted it to be symmetric, you should span from just below the minimum value to just above the maximum value.

btracey: We have the same problem in stat.Histogram. We should make sure that these two functions have the problem in the same direction, or resolve the problem with the same hack.

kortschak: >In principle, if you wanted it to be symmetric, you should span from just below the minimum value to just above the maximum value.

That breaks a different symmetry. Are these not always handled as a set of disjoint intervals [c_n, c_{n+1})? This is how `stat.Histogram` does it. The logic I gave above gives this without depending on the user to provide a carefully prepared set of intervals. (As an aside, now that we are considering this new approach of providing cutoffs, there are reasons why this might be good to provide for `stat.Histogram` or a relative).

Armadilloa16: right, so if the cutoffs are specified by the user its all good -- and no intervals are really involved. The issue is when cutoffs is nil or empty -- in which case ROC defaults to calculating all possible cutoffs, and therein if n y-values are included, y_1, y_2, ..., y_n, then this creates n+1 intervals, where cutoffs in any one of these intervals will produce the same fpr and tpr, so a maximum of n+1 unique (fpr, tpr) pairs. (n-1) of which are of the half-open form [y_k, y_{k+1}) for k = 1, 2, ..., n-1, then there are the two other `special' intervals, (-Inf, y_1) and [y_n, Inf).

roc.go
Armadilloa16: This would produce len(y) == len(fpr), which is wrong -- in order to calculate all possible cutoffs you should have len(y) + 1 == len(fpr). The effect of this change would be to start at (0,0), and then include the first two y values in the next `interval', and then the remainder would be correct.

roc_example_test.go
Armadilloa16: One option would be to just panic if cutoffs is nil or empty, and make people construct cutoffs that would generate all possible values themselves. But I think its nice because that behaviour is often wanted.

kortschak:
What do we want to do with this?

kortschak:
@btracey?

btracey:
I think this is a good change, but it doesn't seem like it's ready to submit immediately. I would suggest we file an issue if someone wants to take the helm

vladimir-ch:
There is already an issue https://github.com/gonum/gonum/issues/708 in the new repository which has been moved there from this one.

I'm closing this PR but please feel free @Armadilloa16 to port this PR to the new repository https://github.com/gonum/gonum

PR #114, opened 2016-07-29
stat: add simple linear regression

kortschak:
@btracey Please take a look.


stat.go
```diff
@@ -648,6 +648,47 @@ func KullbackLeibler(p, q []float64) float64 {
...
+// It returns the values for \hat{\alpha}, \hat{\beta} and the coefficient
+// of determination, r. If origin is true, the regression is forced to pass
+// through the origin.
+func OrdinaryLinearRegression(x, y, weights []float64, origin bool) (alpha, beta, r float64) {
+	w := 1.0
+	yMean := Mean(y, weights)
+	if origin {
+		var x2Sum, xySum float64
+		for i, xi := range x {
+			if weights != nil {
```
sbinet: perhaps a premature optimization, but...
instead of doing this test over and over here (and later, at line 677), wouldn'it be better to create a small `func` literal:

```go
weight := func(i int) float64 {
    return weights[i]
}
if weights == nil {
    weight = func(int) float64 { return 1 }
}
```
and then use that everywhere ?

stat.go
```diff
@@ -648,6 +648,47 @@ func KullbackLeibler(p, q []float64) float64 {
 	return kl
 }
 
+// OrdinaryLinearRegression computes the simple linear regression of x and y.
+// It returns the values for \hat{\alpha}, \hat{\beta} and the coefficient
+// of determination, r. If origin is true, the regression is forced to pass
+// through the origin.
+func OrdinaryLinearRegression(x, y, weights []float64, origin bool) (alpha, beta, r float64) {
```
sbinet: document that `OrdinaryLinearRegression` will panic if `x` and `y` have different lengths.
that test is done explicitly in `Covariance`, when `origin==false` and implicitly when `origin==true` at line 664.
perhaps the test should be done explicitly in `OrdinaryLinearRegression` as well.

document also that `weights` can be `nil` and if not `nil` then it should have the same length than `x` and `y` ?

stat.go
kortschak: I'm not convinced that a closure is likely to be better than a condition; the condition will be well predicted while the closure cannot be inlined. I think this is clearer too.

stat.go
kortschak: Done.

stat.go
```diff
@@ -648,6 +648,71 @@ func KullbackLeibler(p, q []float64) float64 {
 	return kl
 }
 
+// OrdinaryLinearRegression computes the simple linear regression of x and y.
+// It returns the values for \hat{\alpha}, \hat{\beta} and the coefficient
+// of determination, r. If origin is true, the regression is forced to pass
+// through the origin.
+// The lengths of x and y must be equal. If weights is nil then all of the
```
btracey: space in comment?

stat.go
```diff
@@ -648,6 +648,71 @@ func KullbackLeibler(p, q []float64) float64 {
 	return kl
 }
 
+// OrdinaryLinearRegression computes the simple linear regression of x and y.
```
btracey: Do we really need "Ordinary"? Ordinary is in general implied, it's the non-ordinary ones that need an additional moniker.

stat.go
```diff
@@ -648,6 +648,71 @@ func KullbackLeibler(p, q []float64) float64 {
...
+// It returns the values for \hat{\alpha}, \hat{\beta} and the coefficient
```
btracey: We should write out the parameters in the equation. There was a really cool map I saw at one point that showed which constants various parts of the world learned (I learned it as m*x + b).

sbinet:
BTW: should this live in `gonum/stat` ? I was working on a somewhat more general regression functionnality (possibly integrated with gradient descent and `gonum/optimize.Linesearcher`) which I had started to put under `gonum/optimize/fit`...

stat.go
```diff
@@ -648,6 +648,71 @@ func KullbackLeibler(p, q []float64) float64 {
...
+// of determination, r. If origin is true, the regression is forced to pass
+// through the origin.
```
btracey: Here is my rephrasing

````
LinearRegression computes the best-fit line
     y = a*x + b
to the data in x and y with the given weights. If origin is true, the regression is forced to pass through the origin.

Specifically, LinearRegression computes the values of a and b such that the total residual
      \sum_i w[i]*(y[i] - a*x[i] - b)^2
is minimized. If origin is true, then b is forced to be zero.

The lengths of x and y must be equal. If weights is nil then all of the weights are 1. If weights is not nil, then len(x) must equal len(weights).

Comment (btracey) [inline stat.go]: Oops, forgot to add comments about r. I would similarly put the correct words in the first half, and then the equation for r in the second half.

Comment (btracey) [inline stat.go]: Oh, I would also add a comment that mat64.Solve can be used for the same operation in higher dimensions.

**Comment by btracey** [comment]
@sbinet I think this is fine here. It's a very simple code that's easy to understand, and commonly used. For similar reasons, mat64.Solve belongs there, even though it's a special case of linear regression.

I do also agree that we should have a more generalized linear solve. One may want a different loss function (one norm, huber loss), or  do regularization (i.e. Tikhonov). 

In my opinion, these would belong in `gonum/fit` (not a subset of optimize -- regression uses optimization, but it's not a subset of it). However, I would like to see a proposal document before we actually put in code. I believe an ML library should be constructed on a component level, not in the algorithm/silo style that most packages are organized.

File: stat.go [hunk 7]
```diff
@@ -648,6 +648,71 @@ func KullbackLeibler(p, q []float64) float64 {
...
+// The lengths of x and y must be equal. If weights is nil then all of the
+// weights are 1. If weights is not nil, then len(x) must equal len(weights).
```
btracey: It seems better to not have r^2 returned here and just have the separate function
````
RSquared(x, y []float64, a, b float64) float64
````

Reasons:
a) There should be a way to calculate R^2 from a general line, as there are many ways to get a line.
b) Calculating R here doesn't save calculation time as opposed to calculating it in a separate function
c) Given a and b, the functionalities seem orthogonal, and separate functions show that better
d) R^2 isn't the only measure of goodness of fit (just the standard one for lines), and including it here gives R^2 a special significance

File: stat.go (hunk 3 above)
Comment (kortschak) [inline stat.go]: Needed or needed to be deleted? Sorry, I don't understand.

File: stat.go (hunk 4 above)
Comment (kortschak) [inline stat.go]: If there were no weights parameter, I would be happy to do that, but the presence of the weights parameter leads to the possible ambiguity that this does ordinary and weighted least squares linear regression. It doesn't AFAIU.

File: stat.go (hunk 3 above)
Comment (kortschak) [inline stat.go]: Ahh, I see. You are meaning for paras. Note that that is not applied uniformly here. Maybe that should be fixed.

File: stat.go (hunk 7 above)
Comment (kortschak) [inline stat.go]: a) OK. Note that the calculation changes dependent on whether the regression is through the origin or not.

b) There is a duplication of the Mean call in the case that the regression is not through the origin.

d) It is given prominence in R. Normally I would not defer to other languages, but R is the de facto standard for statistical computing, backed by many statisticians.

*If it were* added, it would need to have a weights parameter. The meaning of this is difficult.

Comment (kortschak) [inline stat.go]: Done, but not squashed so the two approaches can be compared.

File: stat.go [hunk 8]
```diff
@@ -648,6 +648,94 @@ func KullbackLeibler(p, q []float64) float64 {
...
+		return alpha, beta
+	}
+
+	beta = Covariance(x, y, weights) / Variance(x, weights)
+	alpha = Mean(y, weights) - beta*Mean(x, weights)
+
+	return alpha, beta
+}
+
+// RSquared returns the coefficient of determination for the best-fit
```
btracey: s/best-fit// (it's any line)

stat.go
```diff
@@ -648,6 +648,94 @@ func KullbackLeibler(p, q []float64) float64 {
 	return kl
 }
 
+// LinearRegression computes the best-fit line
+//  y = alpha*x + beta
```
btracey: This documentation is the opposite of the code. The code computes alpha as the offset and beta as the slope. Personally, I think it's nicer to return the slope first.

stat.go
```diff
@@ -648,6 +648,94 @@ func KullbackLeibler(p, q []float64) float64 {
...
+			x2Sum += w * xi * xi
+		}
+		beta = xySum / x2Sum
+
+		return alpha, beta
+	}
+
+	beta = Covariance(x, y, weights) / Variance(x, weights)
+	alpha = Mean(y, weights) - beta*Mean(x, weights)
+
```
btracey: remove newline?

stat.go
```diff
@@ -648,6 +648,94 @@ func KullbackLeibler(p, q []float64) float64 {
...
+	return alpha, beta
+}
+
+// RSquared returns the coefficient of determination for the best-fit
+// line
+//  y = alpha*x + beta
```
btracey: Same as above, line doesn't match.

stat.go
```diff
@@ -648,6 +648,94 @@ func KullbackLeibler(p, q []float64) float64 {
...
+	}
+
+	w := 1.0
+	if alpha == 0 {
+		var ssr, tot float64
+		for i, xi := range x {
+			if weights != nil {
+				w = weights[i]
+			}
+			fi := beta * xi
```
btracey: This doesn't look right either. Shouldn't the residual be `y[i] - beta * x[i]`, and then at the end be 1 - ssr/tot. If you're trying to use the other formula, it needs to use ymean (which is worse, because then we need two passes over the data).

btracey: Sorry, I was wrong about the yMean thing. This could be right, it's just not obvious to me.

kortschak: It is described [here](https://online.stat.psu.edu/~ajw13/stat501/SpecialTopics/Reg_thru_origin.pdf).

I have changed this though so we are not assuming an alpha of zero means a forced origin point.

stat.go
kortschak: Sorry. Done.

stat.go
kortschak: Done.

stat.go
kortschak: Done.

stat.go
kortschak: Done.

kortschak:
If you want the separate RSquare function, then the two commits should be squashed. Let me know.

stat.go
btracey: Sorry, I didn't realize R^2 was so messed up.

I'm having a tough time removing my statistician hat, so this is a long reply.

In the article you link `Heedlessly applying equation (4) would then result in an implausibly negative (and thus uninterpretable coefficient of determination`. I'm sorry, but I think that's wrong. R^2 is _not_ a goodness of fit metric, at least not directly. It's a measure of variance reduction, or from Wikipedia `the proportion of the variance in the dependent variable that is predictable from the independent variable`. One can have bad models with high R^2, and the correct model with low R^2 (see section 2.1.1 of [Cosma Shalizi's notes](http://www.stat.cmu.edu/~cshalizi/402/lectures/02-truth-about-linear-regression/lecture-02.pdf) ). Since R^2 is a measure of the reduced variance, a negative R^2 has a very clear interpretation -- the fit has actually increased the variance of the data!

There does appear to be debate on the actual definition of R^2. Some define it as `1 - SSres / SStot` (like Matlab), and others define it as `SSreg / SStot`, (like Minitab). It seems clear to me that we should use the first definition (which is eq. 4 in your link).  This definition works for all lines (and really all regressors, linear or not). If it happens to be that `cov(f_i, y_i) = 0` for the regression, (as is true for OLS) then the other definition is a neat fact. If we take the second equation as a definition, then R^2 _only_ makes sense for OLS with an intercept. 

Let's take an extreme example, and consider the predictor `f_i = inf`. With the first definition, `R^2 = -inf`; that's clearly really bad. Under the second definition, R^2 is `inf`. Is that really good? Similarly, there exists a `f(x) = const.` line where `SSreg = SStot`. Thus, I get an R^2 of 1 even though I clearly have a bad predictor. The same flaw does not exist under the residual definition. The only way to have an R^2 of 1 is to make perfect predictions everywhere. (The same flaw, by the way, exists with the definition of 4' in your article).

In short, I think we should define R^2 based on the residuals. The equation is simple, and it's clear what's being computed. It computes something reasonable for all lines, not just the single best-fit line. Finally, the answer is the same in the common case of finding R^2 to the best fit line. Since there does seem to be disagreement on the "true" definition, I feel that we are free to do the reasonable one as long as we're clear about it.

btracey: (If you're reading on email, I updated my comment for clarity and precision)

kortschak: I had a chat with one of the statisticians here and they pointed me to [this](http://stats.stackexchange.com/questions/26176/removal-of-statistically-significant-intercept-term-increases-r2-in-linear-mo/26205#26205). I'll be having a meeting with the professor of stats here in about a week for other things, so I can ask him about it then.

kortschak:
I've pushed a version that splits R₀² from R², adding additional documentation. I think this is the wrong thing to do  because it complexifies the case where client code may wish to try the two definitions programatically. On the other hand it clearly distinguishes the differences.

Note that in your argument with f(x) = c, the point is moot since it cannot have a non-zero alpha and so cannot compute the R₀² in either this version or the previous (one because of compile-time constrains and the other runtime checks on alpha). The case where there is an infinite value breaks the assumptions of the linear model anyway.

kortschak:
I've just had seen the professor of stats and his position is that there are assumptions in the use of the general case that are not met when the line has been regressed through the origin. So this means we need to provide both, either as a function with a flag or as two functions. I prefer the former, but could be convinced.

btracey:
I have been talking to people here as well, and they seem to agree with my rough arguments. I'm hoping to write them up, and maybe you'll find them convincing as well.

One of the main conclusions is that I really don't like calling the two different metrics the same name. For one,  they have very different properties. Normal R^2 is invariant under data translation, R^2_0 is not.  For two, they aren't even "continuous". If you have an OLS line that passes through the origin, it's R^2 and R^2_0 are different numbers. Neither metric is useless, but they don't seem sufficiently similar to switch between them.

Here is what I would propose as a compromise position. We have two functions, LinearRegression, and LinearRegressionOrigin. The former also returns the normal value of R^2, the other returns R^2_0 (and we highlight in LRO that it's NOT a comparable metric). We then also have RSquared, which computes R^2 based on the residual definition, and can take an arbitrary line as input. I can make a full argument that this is a reasonable metric, if you would like. This way someone who wants to compute R^2 based on the reasonable definition for the line through the origin can do so.

It seems reasonable to split the functions up since the algorithm is pretty different in both cases anyway, and it seems like a reasonable compromise position.

kortschak:
Now I'm confused. Why do we now want to return R^2 from the linear regression. That's not what I'm doing now - there is a single LR function and either one or two functions for calculating the coefficient of determination.

btracey:
I'm sorry, I misunderstood your comment. I should have looked at what you actually did.

I like what's there now. The only thing is that RNoughtSquared is NOT maximized by the best-fit line, so it really doesn't make sense to use it with any line except the best fit line. I'm not sure if there's a good way to change the API to make it impossible to compute R0 for any other line (that doesn't remove other problems).

Lastly, I think in the future we may want "Generalized RSquared", which takes in y_hat, y and weights, and computes the same formula as RSquared. Should they be named `RSquared` and `RSquaredGen`, or `RSquaredLinear` and `RSquared`?

kortschak:
>I like what's there now.

I could be convinced that a single signature is not the way to go (i.e. I prefer the previous commit).

>The only thing is that RNoughtSquared is NOT maximized by the best-fit line, so it really doesn't make sense to use it with any line except the best fit line. I'm not sure if there's a good way to change the API to make it impossible to compute R0 for any other line (that doesn't remove other problems).

I don't see that this is possible or necessary.

`RSquaredFrom(estimates, y, weights []float64) float64` for the generalised case.

btracey:
I like RSquaredFrom.

I don't like RSquared and RNoughtSquared being the same function/return argument. They do not seem sufficiently similar to me to call them the same thing, nor to return them in place of one another.

What I meant by "use with a different line", is that right now RNoughtSquared takes in an arbitrary alpha, but it should only be used with the single alpha returned from LinearRegression with origin= true. 

Here is a proposal that may make us both happy

````
LinearRegression(x, y, weights float64) (alpha,beta, rsquared float64)

LinearRegressionOrigin(x, y, weights []float64) (alpha, rnoughtsquared float64)

RSquared(estimates, values, weights []float64) float64
````

I believe this (somewhat) addresses your concern about too many functions. It addresess my concerns that 
a) RNoughtSquared and RSquared are non-overlapping in the API 
b) RNoughtSquared is only computable from the ideal line  
c) RSquared is computable given some other estimator (it's pretty easy to compute the line prediction for other lines, and will be easy to wrap in with `fit` when we have it). 

It's also not so bad since breaking Origin with non-Origin saves approximately zero lines of code -- they really are different algorithms.

If I have misjudged your concerns, let me know.

**Comment by kortschak** [comment]
Where you write alpha, you mean beta.

I honestly think this is taking hand holding to extremes. A determined user can always find a way to use functions incorrectly. It was my understanding that you did not want to tie R² to the linear regression (https://github.com/gonum/stat/pull/114#discussion_r72819306 d)), but this is exactly what is happening here.

Above:

a) This is already true; RNoughtSquared and RSquared have distinct signatures.
b) It prevents calculating R² on an RTO line if that is wanted unless a []float64 is allocated. From what you've written above some people may want this.
c) This is not different between what exists now.

**Comment by btracey** [comment]
So what would you like to see then? I thought you were unhappy with the current signature set

**Comment by kortschak** [comment]
Yeah, I don't particularly like it, but I can get over that. With the additon of the RSquaredFrom it's less troubling than it was without; with only a single pair I felt that a single function with a flag was satisfactory, but the addition of a third form tips the balance.

**Comment by btracey** [comment]
The though of the above post was as follows: The previous API  fits a line for fixed/not fixed, and computes an R^2-like number for origin/fixed. I originally suggested that it should be split into two functions line fitting and R^2 computing. The newer suggestion was that it could be split on the fixed/not fixed axis. 

I had been hoping this was closer to your preference of one signature, while also not doing too much with one function (or at least what I perceive as one function). If you're less happy with the above proposal then as it currently stands, then we don't have to change it.

**Comment by kortschak** [comment]
So are you OK with this current state?

**Comment by btracey** [comment]
The only thing is if we want to allow "fixed intercept" regression. We allow "fixed to zero", but not arbitrary. It's easy to transform between the two problems (just subtract b from y) but I thought I'd mention it. If we did want that, we could use the above scheme but with `LinearRegressionFixed(x, y, weights []float64, intercept float64) (alpha, rnoughtsquared float64)`

File: stat.go [hunk 13]
```diff
@@ -648,6 +648,140 @@ func KullbackLeibler(p, q []float64) float64 {
...
+	}
+
+	w := 1.0
+	yMean := Mean(y, weights)
+	var res, tot, d float64
+	for i, xi := range x {
+		if weights != nil {
+			w = weights[i]
+		}
+		yi := y[i]
```
btracey: Maybe we should just compute yMean on the fly here so it's only one pass over the data?

stat.go
```diff
@@ -648,6 +648,140 @@ func KullbackLeibler(p, q []float64) float64 {
...
+		d = yi - fi
+		res += w * d * d
+		d = yi - yMean
+		tot += w * d * d
+	}
+	return 1 - res/tot
+}
+
+// RSquaredFrom returns the coefficient of determination defined as
+//  R^2 = 1 - \sum_i w[i]*(value[i] - estimate[i])^2 / \sum_i w[i]*(value[i] - mean(values))^2
```
btracey: Switch the order of value and estimate to match the function signature?

kortschak:
Yes, I considered that. I'm not sure how valuable it is.

stat.go
```diff
@@ -648,6 +648,140 @@ func KullbackLeibler(p, q []float64) float64 {
...
+		panic("stat: slice length mismatch")
+	}
+
+	w := 1.0
+	mean := Mean(values, weights)
+	var res, tot, d float64
+	for i, val := range values {
+		if weights != nil {
+			w = weights[i]
+		}
```
btracey: Same here about `mean`?

stat.go
```diff
@@ -648,6 +648,140 @@ func KullbackLeibler(p, q []float64) float64 {
...
+		tot += w * d * d
+	}
+	return 1 - res/tot
+}
+
+// RNoughtSquared returns the coefficient of determination defined as
+//  R₀^2 = \sum_i w[i]*(beta*x[i])^2 / \sum_i w[i]*y[i]^2
+// for the line
+//  y = beta*x
+// and the data in x and y with the given weights.
```
btracey: Could you please add a warning note here that the statistic should only be used for the best-fit line to the data?

stat.go
kortschak: How do we calculate the mean by the time we have reached y[0] without having already iterated over y?

btracey: Sorry, you're right, I'm an idiot.

stat.go
btracey: Nope, my mistake.

btracey:
Code LGTM with the two comment changes.

I'd also be happy if you wanted to switch it to "LinearRegression" / "LinearRegressionFixed" / "RSquared[From]".

stat.go
kortschak: This order is the normal order when it's written with ȳ and ŷ, but for the signature I wanted it to read as "R² from estimates".

btracey: I'm happy with the input order, I just think it's clearer if the description order matches the input order. It doesn't matter that much for clarity in the equation, so it seems easier to just match.

kortschak:
PTAL

btracey:
LGTM

PR #118, opened 2016-08-30
stat: package description

Armadilloa16:
So I was reading the `CovarianceMatrix` function and saw it used `SymOuterK` which in the description says it calculates `s = alpha * x * x'`, which confused me until I saw that it was operating on the transpose of the input. This made me realise that there should be a package-wide convention for how data matrices are treated. This is explicitly included in the comment for `PrincipalComponents` but it seemed reasonable to me that it should just be made a blanket statement to force everything to be consistent. 

We are currently using the more popular convention -- rows as observations -- which is the same as base R uses, but it is not uncommon, particularly in multivariate high-dimensional statistics, to use the other convention -- columns as observations. So I thought it was important to be clear which convention we are using moving forward.

@kortschak @btracey @jonlawlor what do you think?

### DIFF
--- a/stat.go
+++ b/stat.go
@@ -3,6 +3,8 @@
 // license that can be found in the LICENSE file.
 
 // Package stat provides generalized statistical functions.
+// Functions that take a data matrix argument assume that rows correspond to observations
+// and columns to variables.
 package stat
 
 import (


btracey:
I think if we do this, it would probably nice to include more of the package invariants (weights, etc.), and provide pointers to other packages in stat.

Armadilloa16:
Seems reasonable to me, when I have some time would you like me to read through what we currently have in stat and make a suggested list of package invariants? Does it make sense to try and give these consistent variable names? I'm pretty sure weights is called `weights` pretty universally, but data is often given different names: `CovarianceMatrix' calls it `x`, `PrincipleComponents` calls it `a`

btracey:
I'm happy for this to be expanded upon and brought to the new repository.

PR #119, opened 2016-08-30
stat: allow symmetric matrix reuse

kortschak:
Requires https://github.com/gonum/matrix/pull/394 (for any utility gain).

### DIFF
--- a/covariancematrix.go
+++ b/covariancematrix.go
@@ -19,10 +19,10 @@ import (
 // If weights is not nil the weighted covariance of x is calculated. weights
 // must have length equal to the number of rows in input data matrix and
 // must not contain negative elements.
-// If cov is nil, a new matrix with appropriate size will be constructed.
-// If cov is not nil, it must have the same number of columns as the input
-// data matrix, and it will be used as the destination for the covariance
-// data.
+// If cov is not nil it must either be zero-sized or have the same number of
+// columns as the input data matrix. cov will be used as the destination for
+// the covariance data. If cov is nil, a new mat64.SymDense is allocated for
+// the destination.
 func CovarianceMatrix(cov *mat64.SymDense, x mat64.Matrix, weights []float64) *mat64.SymDense {
 	// This is the matrix version of the two-pass algorithm. It doesn't use the
 	// additional floating point error correction that the Covariance function uses
@@ -32,7 +32,7 @@ func CovarianceMatrix(cov *mat64.SymDense, x mat64.Matrix, weights []float64) *m
 
 	if cov == nil {
 		cov = mat64.NewSymDense(c, nil)
-	} else if n := cov.Symmetric(); n != c {
+	} else if n := cov.Symmetric(); n != c && n != 0 {
 		panic(matrix.ErrShape)
 	}
 
@@ -80,10 +80,10 @@ func CovarianceMatrix(cov *mat64.SymDense, x mat64.Matrix, weights []float64) *m
 // If weights is not nil the weighted correlation of x is calculated. weights
 // must have length equal to the number of rows in input data matrix and
 // must not contain negative elements.
-// If corr is nil, a new matrix with appropriate size will be constructed.
-// If corr is not nil, it must have the same number of columns as the input
-// data matrix, and will be used as the destination for the correlation
-// data.
+// If corr is not nil it must either be zero-sized or have the same number of
+// columns as the input data matrix. corr will be used as the destination for
+// the correlation data. If corr is nil, a new mat64.SymDense is allocated for
+// the destination.
 func CorrelationMatrix(corr *mat64.SymDense, x mat64.Matrix, weights []float64) *mat64.SymDense {
 	// This will panic if the sizes don't match, or if weights is the wrong size.
 	corr = CovarianceMatrix(corr, x, weights)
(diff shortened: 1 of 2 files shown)


covariancematrix.go
```diff
@@ -12,15 +12,17 @@ import (
...
 //
-// The weights must have length equal to the number of rows in
-// input data matrix x. If cov is nil, then a new matrix with appropriate size will
-// be constructed. If cov is not nil, it should have the same number of columns as the
-// input data matrix x, and it will be used as the destination for the covariance
-// data. Weights must not be negative.
+// If weights is not nil the weighted covariance of x is calculated. weights
+// must have length equal to the number of rows in input data matrix and
+// must not contain negative elements.
+// If cov is zero sized, a new matrix with appropriate size will be constructed.
```
btracey: This isn't really what it does. If cov is nil, then a new matrix is allocated. If cov is zero-sized, then it is stored into that receiver.

btracey:
I'm not necessarily opposed to this, though I do want to think. The slice arguments are either `nil` or the right size. Does this change also imply all the slices should change? What's the rule for other types?

What I like now is that users should either see `a := function(nil, stuff)` or `function(a, stuff)`. The literal nil is a clear signal that that thing is unused, and our functions panic if `a` is wrong. This PR underminds that. I understand the argument about memory reuse, but maybe this isn't the right solution. Another solution would be something like `func (m *Dense) MakeSize(m,n int)` (with a better name). This would slice Dense to the appropriate size if it is large, and allocate new memory if it isn't. Then the use is

````
sym.Reset()
_, n := data.Dims()
sym.MakeSize(n)
stat.Covariance(sym, data)
````
This keeps the properties of `Covariance` while also allowing data reuse. It's also much clearer where allocations can happen, and where data may no longer be shared.

File: covariancematrix.go (hunk 1 above)
Comment (kortschak) [inline covariancematrix.go]: The language is intentionally thus. The notion of construction is not a Go language concept.

**Comment by kortschak** [comment]
>I'm not necessarily opposed to this, though I do want to think. The slice arguments are either `nil` or the right size.

The approach in this PR is consistent with usage in mat64.

>Does this change also imply all the slices should change?

No.

>What I like now is that users should either see `a := function(nil, stuff)` or `function(a, stuff)`. The literal nil is a clear signal that that thing is unused, and our functions panic if a is wrong. This PR underminds that. I understand the argument about memory reuse, but maybe this isn't the right solution. Another solution would be something like `func (m *Dense) MakeSize(m,n int)` (with a better name). This would slice Dense to the appropriate size if it is large, and allocate new memory if it isn't. Then the use is

```
sym.Reset()
_, n := data.Dims()
sym.MakeSize(n)
stat.Covariance(sym, data)
```

I don't see that this is clearer than

```
sym.Reset()
stat.Covariance(sym, data)
```

>This keeps the properties of Covariance while also allowing data reuse. It's also much clearer where allocations can happen, and where data may no longer be shared.

If we were to do this the Reset call is redundant. It would essentially be an export of the reuseAs (or reuseAsZeroed). The fact that there are two ways this can be called, or that there are two calls that can be chosen argues against it. The receiver knows how it should be used in any given operation, so we tell it that it is OK to work (by Reset()ing it).

File: covariancematrix.go (hunk 1 above)
Comment (btracey) [inline covariancematrix.go]: Sure, but it's also not obvious that `nil` is zero-sized. The operations in mat64 don't allow nil arguments, so I don't think it's obvious that this does. Maybe "If cov is nil or zero sized, a new ..."

**Comment by btracey** [comment]
Okay, I buy it.

File: covariancematrix.go (hunk 1 above)
Comment (kortschak) [inline covariancematrix.go]: I can change the comment, but the proposed start is not correct; a nil and zero sized are different different in their consequences in Go terms, but are the same in their consequences in more abstract conception. So if we distinguish them, we are speaking in Go terms. I'll rewrite it that way - it needs more thought because now there are three cases.

**Comment by kortschak** [comment]
PTAL

File: covariancematrix.go [hunk 2]
```diff
@@ -12,15 +12,17 @@ import (
...
-// The weights must have length equal to the number of rows in
-// input data matrix x. If cov is nil, then a new matrix with appropriate size will
-// be constructed. If cov is not nil, it should have the same number of columns as the
-// input data matrix x, and it will be used as the destination for the covariance
-// data. Weights must not be negative.
+// If weights is not nil the weighted covariance of x is calculated. weights
+// must have length equal to the number of rows in input data matrix and
+// must not contain negative elements.
+// If cov is not nil it must either be zero-sized or have the same number of
+// columns as the input data matrix, cov will be used as the destination for
```
btracey: s/,/. (or s/,/, and)

covariancematrix.go
```diff
@@ -72,14 +74,16 @@ func CovarianceMatrix(cov *mat64.SymDense, x mat64.Matrix, weights []float64) *m
...
-// The weights must have length equal to the number of rows in
-// input data matrix x. If corr is nil, then a new matrix with appropriate size will
-// be constructed. If corr is not nil, it should have the same number of columns
-// as the input data matrix x, and it will be used as the destination for the
-// correlation data. Weights must not be negative.
+// If weights is not nil the weighted correlation of x is calculated. weights
+// must have length equal to the number of rows in input data matrix and
+// must not contain negative elements.
+// If corr is not nil it must either be zero-sized or have the same number of
+// columns as the input data matrix, corr will be used as the destination for
```
btracey: same as above.

btracey:
LGTM with minor fix.

PR #120, opened 2016-08-30
Add Gamma distribution and tests

btracey:


distuv/gamma.go
```diff
@@ -0,0 +1,155 @@
...
+
+// Prob computes the value of the probability density function at x.
+func (g Gamma) Prob(x float64) float64 {
+	return math.Exp(g.LogProb(x))
+}
+
+// Rand returns a random sample drawn from the distribution.
+func (g Gamma) Rand() float64 {
+	if g.Beta <= 0 {
+		panic("gamma: beta < 0")
```
kortschak: Comment above that these constraints exist.

distuv/gamma.go
```diff
@@ -0,0 +1,155 @@
...
+		//  Statistical Software 55.1 (2013): 1-17.
+		// Algorithm 2.
+		umax := math.Pow(a/math.E, a/2)
+		vmin := -2 / math.E
+		vmax := 2 * a / math.E / (math.E - a)
+		for {
+			u := umax * unifrnd()
+			t := (unifrnd()*(vmax-vmin) + vmin) / u
+			t1 := math.Exp(t / a)
+			if 2*math.Log(u) > t-t1 {
```
kortschak: It seems to me that this would be clearer if the last four lines of the loop were dropped out of the loop and this condition were inverted and changed to a break, of the last four lines were included in this inverted condition.

distuv/gamma.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+	case a > 1:
+		// Generate using
+		//  Martino, Luca, and David Luengo. "Extremely efficient generation of
+		//  Gamma random variables for α>= 1." arXiv preprint arXiv:1304.3800 (2013).
+		ap := math.Floor(a)
+		bp := b * (ap - 1) / (a - 1)
+		lkp := (ap - a) + (a-ap)*(a-1)/b
+		if ap == 1 {
+			bp = b / a
```
kortschak: The paper says this happens when a < 2.

distuv/gamma.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+			return t / a / b
+		}
+	case a > 1:
+		// Generate using
+		//  Martino, Luca, and David Luengo. "Extremely efficient generation of
+		//  Gamma random variables for α>= 1." arXiv preprint arXiv:1304.3800 (2013).
+		ap := math.Floor(a)
+		bp := b * (ap - 1) / (a - 1)
+		lkp := (ap - a) + (a-ap)*(a-1)/b
```
kortschak: Put this in an else block below.

distuv/gamma.go
btracey: Done.

distuv/gamma.go
btracey: Done.

distuv/gamma.go
btracey: Done.

distuv/gamma.go
btracey: ap = math.Floor(a) so it has to be an integer, and a >= 1. I can change it if you'd like, but they're identical.

btracey:
PTAL

distuv/gamma.go
kortschak: Sorry, yes. Maybe just add a comment in the conditional block for slow people like me.

distuv/gamma.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// Prob computes the value of the probability density function at x.
+func (g Gamma) Prob(x float64) float64 {
+	return math.Exp(g.LogProb(x))
+}
+
+// Rand returns a random sample drawn from the distribution.
+//
+// Rand panics if either alpha or beta is <= 0.
```
kortschak: Can you put these constraints in the struct field docs as well.

distuv/gamma.go
```diff
@@ -0,0 +1,160 @@
...
+//
+// For more information, see https://en.wikipedia.org/wiki/Gamma_distribution
+type Gamma struct {
+	// Alpha is the shape parameter of the distribution. If Shape == 1, this is
+	// equivalent to an exponential distribution.
+	Alpha float64
+	// Beta is the rate parameter of the distribution. If Rate == 2, this is
+	// equivalent to a Chi-Squared distribution.
+	Beta   float64
+	Source *rand.Rand
```
kortschak: A blank between Beta and Source would be make the grouping clearer.

kortschak:
LGTM with minor comments addressed.

distuv/gamma.go
btracey: Done.

distuv/gamma.go
btracey: Done.

PR #122, opened 2016-09-01
Add Dirichlet distribution and tests

btracey:


btracey:
PTAL @kortschak 

I'd like to hold off on making this work for zero-sized matrices, and then fix it in the next PR with Normal and whatever else needs it. However, I don't even see how to do that with the mat64 API. CovarianceMatrix, for example, just needs to set values, but as far as I can tell there is no way to set the matrix size by hand.

kortschak:
Grow and View?

distmv/dirichlet.go
```diff
@@ -0,0 +1,142 @@
...
+// return, the value at element {i, j} of the covariance matrix is equal to
+// the covariance of the i^th and j^th variables.
+//  covariance(i, j) = E[(x_i - E[x_i])(x_j - E[x_j])]
+// If the input matrix is nil a new matrix is allocated, otherwise the result
+// is stored in-place into the input.
+func (d *Dirichlet) CovarianceMatrix(cov *mat64.SymDense) *mat64.SymDense {
+	if cov == nil {
+		cov = mat64.NewSymDense(d.Dim(), nil)
+	} else if cov.Symmetric() != d.dim {
+		panic("normal: input matrix size mismatch")
```
kortschak: Ahh. I see what you mean here. This needs thought. We seem to need a cognate for ViewSquare.

kortschak: I was just writing the issue and had a look. You added a GrowSquare which will do what you want here, no?

`if cov.Symmetric() == 0 { cov.GrowSquare(0, d.dim) }`?

btracey: I think the idea should work, though that, as written, doesn't work, right? We have to update the actual pointer itself.

````
if cov.Symmetric() == 0 {
     m := cov.GrowSquare(0, d.dim).(*mat64.SymDense) 
    *cov = *m
}

````

Comment (kortschak) [inline distmv/dirichlet.go]: You are right.

Comment (kortschak) [inline distmv/dirichlet.go]: Though is can be hideously `*cov = *(cov.GrowSquare(0, d.dim).(*mat64.SymDense))`.

Comment (btracey) [inline distmv/dirichlet.go]: Should I fix this case here, or with Normal?

Comment (kortschak) [inline distmv/dirichlet.go]: Sorry, I don't understand.

**Comment by btracey** [comment]
Okay, adding allowing a zero matrix.

File: distmv/dirichlet.go [hunk 2]
```diff
@@ -0,0 +1,144 @@
...
+// is stored in-place into the input.
+func (d *Dirichlet) CovarianceMatrix(cov *mat64.SymDense) *mat64.SymDense {
+	if cov == nil {
+		cov = mat64.NewSymDense(d.Dim(), nil)
+	} else if cov.Symmetric() == 0 {
+		*cov = *(cov.GrowSquare(d.dim).(*mat64.SymDense))
+	} else if cov.Symmetric() != d.dim {
+		panic("normal: input matrix size mismatch")
+	}
+	den := d.sumAlpha * d.sumAlpha * (d.sumAlpha + 1)
```
kortschak: Can this be `scale := 1 / d.sumAlpha * d.sumAlpha * (d.sumAlpha + 1)` and use multiplication below.

PR #123, opened 2016-09-05
stat: add canonical correlation analysis function

Armadilloa16:
PTAL

cca.go contains the main addition -- `CanonicalCorrelations` --  and everything else is ancillary to that. 
The idea is to do Canonical Correlation Analysis (CCA) via singular value decomposition. 

Some notes:
- Included are some changes to PrincipalComponents, as the same svdFactorizeCentered is done in a couple of places now its moved into a function.
- I wasn't sure how much detail to go into in the documentation and/or Example, so I tried to be consistent in terminology but please suggest improvements -- @vladimir-ch, @kortschak mentioned to me that you might be particularly helpful in this regard?
- I changed my mind many times about what to return, and ended up with a long list of outputs. The reason I opted to return the left and right eigenvectors of the canonical correlation matrix instead of the arguably more popular canonical variable transforms is because it takes less work, and if the user wants the transforms they can easily produce them using the returned svd's -- and I include a demonstration of how to do this at the end of the Example. I also played with the idea of just returning the SVD of the canonical correlation matrix (which just constitutes the first three currently returned values) but when making the example realised you are always going to need to unpack it anyway, so I might as well return the unpacked components.
- I also added a couple of datasets: Boston and Car, primarily for testing purposes as they are used in examples in the textbook I was working off (details in the comment), but these could potentially be used elsewhere in other examples/ tests as well.

Thanks for all the help and patience answering my silly questions while I was working on this @kortschak.


cca.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	"errors"
+	"math"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix"
+	"github.com/gonum/matrix/mat64"
+)
+
+// CanonicalCorrelations returns the `canonical' correlations between
```
kortschak: Remove quotes.

cca.go
```diff
@@ -0,0 +1,122 @@
...
+// the corresponding pairs of canonical variables -- described by the
```
kortschak: We can use unicode, so if you want an m-dash, use "—". However, a comma works fine.

cca.go
```diff
@@ -0,0 +1,122 @@
...
+// left and right eigenvectors of the canonical correlation matrix,
+// pvecs and qvecs respectively. The canonical correlation matrix is
```
kortschak: The description of left/right, pvecs/qvecs is too dense, harming the parsing.

cca.go
```diff
@@ -0,0 +1,122 @@
...
+// calculated using the singular value decompositions of the centered
+// inputs, and these singular value decompositions, xsvd and ysvd, are
```
kortschak: Avoid ", and" in these kinds of cases. Two sentences works better; "These singular...".

cca.go
```diff
@@ -0,0 +1,122 @@
...
+// also returned.
+//
+// The slice weights is used to weight the observations. If weights is nil,
+// each weight is considered to have a value of one, otherwise the length of
+// weights must match the number of observations or CanonicalCorrelations will
+// panic.
+//
+// Canonical Correlations will panic if the inputs x and y do not have the same
```
kortschak: s/Canonical Correlations/CanonicalCorrelations/

cca.go
```diff
@@ -0,0 +1,122 @@
...
+// number of rows, and their columns should be interepretable as two different
+// sets of measurements (variables) on the same observations (each row representing
```
kortschak: The second parenthesis should be a new sentence.

cca.go
```diff
@@ -0,0 +1,122 @@
...
+// an observation).
+//
+// The sphered data are Xc * Sx^(-1/2) and Yc * Sy^(-1/2) respectively, and the
```
kortschak: Sphering has not yet been introduced.

cca.go
```diff
@@ -0,0 +1,122 @@
...
+// correlation matrix between the sphered data is the canonical correlation matrix,
+// Sx^(-1/2) * Sxy * Sy^(-1/2). In cases where S^(-1/2) is ambiguous, S^(-1/2) is
+// taken to be E * D^(-1/2) * E' where S can be eigendecomposed as S = E * D * E'.
```
kortschak: Introduce S earlier.

cca.go
```diff
@@ -0,0 +1,122 @@
...
+// Xc and Yc denote the centered input data matrices x and y (column means subtracted
+// from each column), Sx and Sy denote the sample covariance matrices within x and y
+// respectively, and Sxy denotes the covariance matrix between x and y.
```
kortschak: I think in general the ordering of this is wrong. I understand why, you're writing this as a maths text, so you have the eqn up front with the parts explained. If you want to have it in that order, we usually have the eqn and then in \<pre\> blocks (one extra space indent in the comment) we have the parts so they are easy to find.

boston_data_test.go
```diff
@@ -0,0 +1,530 @@
...
+
+package stat
+
+import "github.com/gonum/matrix/mat64"
+
+// Boston Housing Data of Harrison and Rubinfeld (1978)
+// http://dx.doi.org/10.1016/0095-0696(78)90006-2
+// http://lib.stat.cmu.edu/datasets/boston
+// Columns are: Per capita crime rate by town,
+//              Proportion of non-retail business acres per town,
```
kortschak: This doesn't show up in the godoc, so it won't be a problem, but rather than this (which will be wrong if it moves somewhere where it does get rendered), use something along the lines of

```
// Columns are;
//  per capita crime rate by town,
//  proportion of non-retail business acres per town,
//  nitric oxide concentration (parts per 10 million),
//  weighted distances to Boston employment centres,
//  index of accessibility to radial highways,
//  pupil-teacher ratio by town,
//  proportion of blacks by town,
//  average number of rooms per dwelling,
//  proportion of owner-occupied units built prior to 1940,
//  full-value property-tax rate per $10000,
//  median value of owner-occupied homes in $1000s.
```

(also note that this is now a sentence).

cca.go
```diff
@@ -0,0 +1,122 @@
...
+//
+// The first pair of left and right eigenvectors of the canonical correlation matrix
+// can be interpreted as directions into which the respective sphered data can be
+// projected such that the correlation between the two projections is maximised.
+// The second pair and onwards solve the same optimisation but under the constraint
```
kortschak: Unfortunately, American spelling on public-facing docs please.

cca.go
```diff
@@ -0,0 +1,122 @@
...
+// that they are uncorrelated to previous projections.
+//
+// More details can be found at https://en.wikipedia.org/wiki/Canonical_correlation
+// or in Chapter 3 of
+// Koch, Inge. Analysis of multivariate and high-dimensional data.
+// Vol. 32. Cambridge University Press, 2013. ISBN: 9780521887939
+//
+// If one of the necessary singular value decompositions is not possible, an error
+// is returned as err and all other outputs are returned as nil.
```
kortschak: This last comment is not necessary.

cca.go
```diff
@@ -0,0 +1,122 @@
...
+	n, _ := x.Dims()
+	yn, _ := y.Dims()
+	if n != yn {
+		panic("stat: unequal number of observations")
+	}
+	if weights != nil && len(weights) != n {
+		panic("stat: len(weights) != observations")
+	}
+
+	var ok bool
```
kortschak: Delete

cca.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// Centre and factorize x and y.
+	xsvd, ok = svdFactorizeCentered(x, weights)
```
kortschak: `xsvd, ok := svdF...`

cca.go
```diff
@@ -0,0 +1,122 @@
...
+	if !ok {
+		return nil, nil, nil, nil, nil, errors.New("stat: failed to factorize x")
+	}
+	ysvd, ok = svdFactorizeCentered(y, weights)
```
kortschak: `ysvd, ok = svdF...`

cca.go
```diff
@@ -0,0 +1,122 @@
...
+	xu.UFromSVD(xsvd)
+	xv.VFromSVD(xsvd)
+	yu.UFromSVD(ysvd)
+	yv.VFromSVD(ysvd)
+
+	// Calculate and factorise the canonical correlation matrix.
+	var ccor mat64.Dense
+	ccor.Product(&xv, xu.T(), &yu, yv.T())
+	var csvd mat64.SVD
+	if ok := csvd.Factorize(&ccor, svdKind(ccor.Dims())); !ok {
```
kortschak:
```
ok = csvd.Factorize(&ccor, svdKind(ccor.Dims()))
if !ok {
```

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"fmt"
+	"log"
+	"math"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix/mat64"
+	"github.com/gonum/stat"
+)
+
+// Helper for getting a View of a SymDense.
```
kortschak: `// symView is a helper for getting a View of a SymDense.`

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+type symView struct {
+	sym        *mat64.SymDense
+	i, j, r, c int
```
kortschak: Add a line between these.

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	// The columns (variables) of xdata are, in order:
+	//     Per capita crime rate by town,
+	//     Proportion of non-retail business acres per town,
+	//     Nitric oxide concentration (parts per 10 million),
+	//     Weighted distances to Boston employment centres,
+	//     Index of accessibility to radial highways,
+	//     Pupil-teacher ratio by town, and
+	//     Proportion of blacks by town.
+	xdata := stat.BostonData.View(0, 0, n, xd)
+	// The columns (variables) of ydata are, in order:
```
kortschak: New line.

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	// BostonData is the Boston Housing Data of Harrison and Rubinfeld (1978)
+	n, _ := stat.BostonData.Dims()
+	var xd, yd = 7, 4
+	// The variables (columns) of bostonData can be partitioned into two sets:
+	// those that deal with environmental/ social variables (xdata), and those
+	// that contain information regarding the individual (ydata).
+	// Because the variables can be naturally partitioned in this way, these
+	// data are appropriate for Canonical Correlation Analysis (CCA).
+	// The columns (variables) of xdata are, in order:
+	//     Per capita crime rate by town,
```
kortschak: No leading caps; these are a sentence. Replace ":" with ";".

Also below.

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+func (s symView) T() mat64.Matrix { return mat64.Transpose{s} }
+
+func ExampleCanonicalCorrelations() {
+	// This example is directly analogous to Example 3.5 on page 87 of
+	// Koch, Inge. Analysis of multivariate and high-dimensional data.
+	// Vol. 32. Cambridge University Press, 2013. ISBN: 9780521887939
+
+	// BostonData is the Boston Housing Data of Harrison and Rubinfeld (1978)
+	n, _ := stat.BostonData.Dims()
```
kortschak: This is unfortunately misleading. In the first instance I would have the Boston data as an unexported value in `stat_test`. If it becomes important for other tests (non-example) in `stat` to use it, then export it from `stat` and have an unexported `bostonData` aliasing `stat.BostonData` (still provided by a `*_test.go` file). The rationale for this is that when people see `stat.BostonData` they will think it's available; it's not.

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	// For comparison, calculate the correlation matrix for the original data
+	var cor, cov mat64.SymDense
+	stat.CovarianceMatrix(&cov, stat.BostonData, nil)
+	stat.CorrelationMatrix(&cor, stat.BostonData, nil)
+
+	// Extract just those correlations that are between xdata and ydata.
+	var corRaw = symView{sym: &cor, i: 0, j: 7, r: 7, c: 4}
+
+	// Note that the strongest correlation between individual
+	// variables is 0.91 between the 5 th variable of xdata (index of
```
kortschak: s/5 th/5th/

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	// accessibility to radial highways) and the 3 rd variable of ydata
```
kortschak: s/3 rd/3rd/

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	// stronger than the greatest correlation in the sphered data.
+	fmt.Printf("\n\nccors = %.4f", ccors)
+
+	// Left and right eigenvectors of the canonical correlation matrix.
+	fmt.Printf("\n\npVecs = %.4f", mat64.Formatted(pVecs, mat64.Prefix("        ")))
+	fmt.Printf("\n\nqVecs = %.4f", mat64.Formatted(qVecs, mat64.Prefix("        ")))
+
+	// Canonical Correlation Transforms. These can be useful as they represent the
+	// canonical variables as linear combinations of the original variables.
+	// Note that the R function cancor returns xcoef and ycoef which are the
```
kortschak: Remove this sentence.

cca.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	var pv, qv mat64.Dense
+	pv.UFromSVD(&csvd)
+	qv.VFromSVD(&csvd)
+	corrs = csvd.Values(nil)
+
+	return corrs, &pv, &qv, xsvd, ysvd, nil
+}
+
+func svdKind(n, d int) matrix.SVDKind {
```
kortschak: It seems odd that these live here now. I have no suggestion.

cca.go
Armadilloa16: I removed these parentheses, I think they where probably redundant anyway.

cca.go
Armadilloa16: I had a go at re-writing this, seperating into more sentences addressing your other comments and attempting to address this one.

cca.go
Armadilloa16: Is this what you mean? Funny, I had it the other way around earlier, with the notation first. I could always change it back I don't really mind either way that much.

cca.go
Armadilloa16: I think this is needed because xsvd, ok = svdFactorizeCentered(... 
does not use `:=` this is because xsvd is already initialised because its defined as the name of the output variable to be returned. Should I rename it in order to remove this?

cca_example_test.go
Armadilloa16: Ok but this is a style thing I don't understand -- is it differentiating between different types of fields?

kortschak: Pragmatism; it doesn't push `*mat64.SymDense` so far to the right.

cca_example_test.go
Armadilloa16: That makes sense to me. At the moment I am using it in cca_test.go, which is in package stat, same goes for the car data of car_data_test.go. Does that mean I should add a data_test.go or something introducing aliases for stat.BostonData to avoid this confusing? I don't see an alternative unless I remove the tests that use it.

cca.go
Armadilloa16: If we wanted to export them they could go with SVD.Factorise but we probably don't want to export them. They could go in pca.go but that doesn't make too much difference really. In principle we could merge pca.go and cca.go into one file but what would it be called? A minor point really, I suppose.

Armadilloa16:
Ok all the comments I haven't replied too I just implemented as you suggested, thanks @kortschak

cca_example_test.go
kortschak: There are two ways to do it, the way you are suggesting (though with `export_test.go`) or to make `cca_test.go` be in `stat_test` rather than in `stat`.

cca.go
kortschak: I'd put both cca and pca in the same file. Leave the tests separate.

cca_example_test.go
Armadilloa16: ooooh manipulating go fmt nice

cca_example_test.go
Armadilloa16: Ok yeah I think I like putting `cca_test.go` into `stat_test` better, so thats what I did, but then that required me to copy `func approxEqual` from `pca_test.go`, maybe this really should be somewhere central? Or I could put `pca_test.go` into `stat_test` as well? We should probably try to be consistent about if tests are in the same package, or in a _test package no?

cca.go
Armadilloa16: What would you call the file that has both in it? Maybe `canonicalmvmethods.go` for canonical multivariate methods or something? as PCA is a "canonical" method in a similar sense to the way in which the word is used in CCA. So ugly though.

kortschak: `pca_cca.go`

boston_data_test.go
```diff
@@ -0,0 +1,531 @@
...
+
+package stat_test
+
+import "github.com/gonum/matrix/mat64"
+
+// Boston Housing Data of Harrison and Rubinfeld (1978)
+// http://dx.doi.org/10.1016/0095-0696(78)90006-2
+// http://lib.stat.cmu.edu/datasets/boston
+// Columns are;
+//  Per capita crime rate by town,
```
kortschak: No uppercase in the middle of sentences. Wir nicht sprecken Deutsch.

car_data_test.go
```diff
@@ -0,0 +1,406 @@
...
+// license that can be found in the LICENSE file.
+
+package stat_test
+
+import "github.com/gonum/matrix/mat64"
+
+// ASA Car Exposition Data of Ramos and Donoho (1983)
+// http://lib.stat.cmu.edu/datasets/cars.desc
+// http://lib.stat.cmu.edu/datasets/cars.data
+// Columns are: Displacement, Horsepower, Weight, Acceleration, MPG
```
kortschak: Same here. Full stop at end of sentence.

cca.go
```diff
@@ -0,0 +1,118 @@
...
+	"math"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix"
+	"github.com/gonum/matrix/mat64"
+)
+
+// CanonicalCorrelations returns the canonical correlations, corr,
+// between the corresponding pairs of canonical variables. Canonical
+// variables can be obtained by projecting the sphered data into the
```
kortschak: Sphered is not yet defined.

cca.go
```diff
@@ -0,0 +1,118 @@
...
+// panic.
+//
+// CanonicalCorrelations will panic if the inputs x and y do not have the same
+// number of rows, and their columns should be interepretable as two different
+// sets of measurements on the same observations.
+//
+// The sphered data are Xc * Sx^(-1/2) and Yc * Sy^(-1/2) respectively, and the
+// correlation matrix between the sphered data is the canonical correlation matrix,
+// Sx^(-1/2) * Sxy * Sy^(-1/2).
+// 	In cases where S^(-1/2) is ambiguous, S^(-1/2) is taken to be E * D^(-1/2) * E'
```
kortschak: I'm confused about what you're trying to do with the godoc here.

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	//  Index of accessibility to radial highways,
+	//  Pupil-teacher ratio by town, and
+	//  Proportion of blacks by town.
+	xdata := bostonData.View(0, 0, n, xd)
+
+	// The columns (variables) of ydata are, in order;
+	//  Average number of rooms per dwelling,
+	//  Proportion of owner-occupied units built prior to 1940,
+	//  Full-value property-tax rate per $10000, and
+	//  Median value of owner-occupied homes in $1000s
```
kortschak: Full stop.

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	// bostonData is the Boston Housing Data of Harrison and Rubinfeld (1978)
+	n, _ := bostonData.Dims()
+	var xd, yd = 7, 4
+	// The variables (columns) of bostonData can be partitioned into two sets:
+	// those that deal with environmental/ social variables (xdata), and those
+	// that contain information regarding the individual (ydata).
+	// Because the variables can be naturally partitioned in this way, these
+	// data are appropriate for canonical correlation analysis.
+	// The columns (variables) of xdata are, in order;
+	//  Per capita crime rate by town,
```
kortschak: No capitals in a sentence here and below.

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+func ExampleCanonicalCorrelations() {
+	// This example is directly analogous to Example 3.5 on page 87 of
+	// Koch, Inge. Analysis of multivariate and high-dimensional data.
+	// Vol. 32. Cambridge University Press, 2013. ISBN: 9780521887939
+
+	// bostonData is the Boston Housing Data of Harrison and Rubinfeld (1978)
+	n, _ := bostonData.Dims()
+	var xd, yd = 7, 4
+	// The variables (columns) of bostonData can be partitioned into two sets:
+	// those that deal with environmental/ social variables (xdata), and those
```
kortschak: `s|/ |/|`

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	// Because the variables can be naturally partitioned in this way, these
+	// data are appropriate for canonical correlation analysis.
+	// The columns (variables) of xdata are, in order;
+	//  Per capita crime rate by town,
+	//  Proportion of non-retail business acres per town,
+	//  Nitric oxide concentration (parts per 10 million),
+	//  Weighted distances to Boston employment centres,
+	//  Index of accessibility to radial highways,
+	//  Pupil-teacher ratio by town, and
+	//  Proportion of blacks by town.
```
kortschak: Can I get input from someone with American sensibilities on the use of "blacks"?

cca.go
Armadilloa16: Done.

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	// omitting those within xdata and within ydata.
+	var covRaw = symView{sym: &cov, i: 0, j: xd, r: xd, c: yd}
+
+	// Calculate the canonical correlations.
+	ccors, pVecs, qVecs, xsvd, ysvd, err := stat.CanonicalCorrelations(xdata, ydata, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Unpack xsvd and ysvd
```
kortschak: Full stop.

boston_data_test.go
Armadilloa16: If we were, it would still be wrong because its not a noun. :P

cca.go
Armadilloa16: Same with canonical variables and canonical correlation matrix, I'm not sure how to fix this other than to add a sentence explaining that these ideas will be introduced below.

cca.go
Armadilloa16: Honestly, me too. Maybe I'll just re-arrange it to have the notation first and equations second.

cca_example_test.go
Armadilloa16: To quote Koch, I. "The variable names in the table are those of Harrison and Rubinfeld, rather than names I would have chosen."

Armadilloa16: Could add that in the explanation though.

cca_example_test.go
kortschak: I'm thinking that the best thing to do is export `BostonData` and `CarData` in a test file, but use unexported versions in the examples so that they are not seen. This prevents the func duplication.

boston_data_test.go
kortschak: Sehr schlau!

cca_example_test.go
kortschak: We are allowed to change language, but wait for someone else's input.

btracey:
I'm on vacation at the moment and can't do a useful review, so sorry. However, my first impression is that this strikes me as the wrong level of abstraction. I'm not sure if the right answer is to return less (maybe only coors), or do less (take in the svds as input),  or return a focused type. In any event, these kinds of functions that take in something general and return a whole bunch of things tend not to be not that easy to use or composable with other functions. At a first impression, it seems like maybe it should accept the svds of the data. Then it's clear how to use the function, but here we can avoid the error checking and all the extra returns (because they can be gotten from the svd). Just a thought.

btracey:
An addendum: maybe stat could have a "CenteredSVD" function/type that can be an input to this and pca

cca_example_test.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	// Unpack xsvd and ysvd.
+	var xv, yv mat64.Dense
+	xv.VFromSVD(xsvd)
+	yv.VFromSVD(ysvd)
+	xs := xsvd.Values(nil)
+	ys := ysvd.Values(nil)
+	var xvs, yvs mat64.Dense
+	xvs.Clone(&xv)
```
kortschak: Why are these clone operations necessary?

kortschak:
Funny, @Armadilloa16 was suggesting a centred SVD function. I think that might work; it would be worth seeing what the example looks like with that change. The signature would become `CannonicalCorrelations(xsvd, ysvd *mat64.SVD) (corrs []float64, pvecs, qvecs *mat64.Dense, ok bool)`. It does allow people to incorrectly weight things though.

Alternatively, a `CC struct { x, y, c *mat64.SVD }` type from `CanonicalCorrelations(x, y mat64.Matrix, weights []float64) (CC, error)` with `Corrs([]float64) []float64`, `Vectors(p, q *mat64.Dense) (pvecs, qvecs *mat64.Dense)` and `???() (xsvd, ysvd *mat64.SVD)`. I think I prefer this as it prevents the inevitable 6 lines of preamble for each call to `CannonicalCorrelations`, and significantly, prevents issues being raised when people differentially weight their input SVDs or have otherwise mismatched inputs.

Armadilloa16:
No worries hope you enjoy your vacation @btracey! 

I like the idea of having CenteredSVD function, and giving CanonicalCorrelations the SVD's of your data. 
I have two additional comments on that,
 - This would basically make PrincipalComponents redundant, as PrincipalComponents as it stands only does two things in addition to CenteredSVD; 1. it transforms the singular values into variances (simple enough) and 2. it only keeps the first min(n,d) vectors. I have issues with this second point anyway --- it can be min(n-1,d) because the data is centered and it could be made even more strict if we use rank instead (might need to set an epsilon for singular values close to zero though). I actually think this restriction should be part of SVD.Factorise / CenteredSVD (min(n,d), min(n-1,d) respectively) if it isn't already as it is completely general but that is probably a different discussion.
 - There is a relationship between the output of CenteredSVD and CovarianceMatrix/CorrelationMatrix which should probably be mentioned somewhere in the context of `stat`. If you have got the output of CenteredSVD already then you can probably shortcut some of the calculation for `CovarianceMatrix`/`CorrelationMatrix`

Now I'm thinking about this more the question is really where should the abstraction be. What we are suggestion with CenteredSVD essentially removes almost all the abstraction, is that what we want?

cca_example_test.go
Armadilloa16: Because to calculate the canonical correlation matrix `corSph` below we need to multiply `xv` and `xvs` -- is there a better way to do this?

kortschak: You mean [here](https://github.com/gonum/stat/pull/123/files#diff-8577f2a2f8d5735b4a8e218e8feac0faR119), [here](https://github.com/gonum/stat/pull/123/files#diff-8577f2a2f8d5735b4a8e218e8feac0faR136) and [here](https://github.com/gonum/stat/pull/123/files#diff-8577f2a2f8d5735b4a8e218e8feac0faR138)? Just `corSph.Product(&xv, xv.T(), covRaw, &yv, yv.T())`, `phiVs.Product(&xv, xv.T(), pVecs)` and `phiVs.Product(&xv, xv.T(), pVecs)`.

Armadilloa16: Because I am scaling `xvs` and `yvs` so they are no longer equal to `xv` and `yv` by the time those multiplications happen.

kortschak: Ahh, sorry. Ignore me sometimes.

Armadilloa16: `xvs` and `yvs` are being modified (columns scaled) so when the multiplication occurs they are no longer equal to `xv` and `yv`

Armadilloa16: Actually, there is a good point here though -- I could do this by scaling `xv` and `yv` by `math.Sqrt(1/xs[j])` and `math.Sqrt(1/y[j]` and then do what you suggested -- removing the need for `xvs` and `yvs` and maybe making the computation cheaper? and we already have a dependancy on `math` so it doesn't add that just for this. Is that a good idea?

kortschak: If it looks nicer, do it.

Armadilloa16: Looks very similar, just doesn't require `xvs` and `yvs` to be allocated

kortschak: Allocations are ugly.

Armadilloa16:
I just implemented the CC type version, looks better (particularly the example). I still need to update the comments and adjust the tests (I want to test psi and phi instead of xsvd and ysvd, this should also make `cca_test` much more readable), and I'll have a go at those two things tomorrow.

Armadilloa16:
Ok when you get a chance PTAL @kortschak -- this is the type CC idea. If you have a suggestion for a better name for VectorsX and VectorsY that would be good (maybe left and right instead of X and Y urgh I dunno). Also the comments/ documentation probably needs abit more work, but the basic outline is there and it seems to work nicely -- the example is much neater with all this stuff abstracted away which I think is a plus. `cca_test.go` is also much easier to read now I removed the checks for xsvd and ysvd and replaced them with tests for phi and psi, which isn't outward facing but still is much nicer I think.

kortschak:
`Left` and `Right` SGTM

pca_cca.go
```diff
@@ -0,0 +1,239 @@
...
+	} else {
+		f = 1 / (floats.Sum(weights) - 1)
+	}
+	for i, v := range vars {
+		vars[i] = f * v * v
+	}
+	return vecs, vars, true
+}
+
+// scaleColsReciSqrt is a helper for scaling the columns of cols
```
kortschak: Move this down. As a general principle have helpers follow calling funcs.

pca_cca.go
```diff
@@ -0,0 +1,239 @@
...
+	for j := 0; j < d; j++ {
+		mat64.Col(col, j, cols)
+		floats.Scale(math.Sqrt(1/vals[j]), col)
+		cols.SetCol(j, col)
+	}
+}
+
+// CC holds the relevant SVDs for canonical correlation analysis
+type CC struct {
+	n int
```
kortschak: n needs a better name or a comment, or both.

pca_cca.go
```diff
@@ -0,0 +1,239 @@
...
+
+// Corrs returns the canonical correlations in dest.
+func (c CC) Corrs(dest []float64) []float64 {
+	return c.c.Values(dest)
+}
+
+// VectorsX returns the left-eigenvectors of the canonical correlation
+// matrix if spheredSpace is true, if spheredSpace is false it returns
+// these eigenvectors back-transformed to the original data space.
+func (c CC) VectorsX(spheredSpace bool) (vecs *mat64.Dense) {
```
kortschak: Do we want a dst param?

pca_cca.go
```diff
@@ -0,0 +1,239 @@
...
+
+	var pv mat64.Dense
```
kortschak: This func can be factored out to a helper to reduce the duplication.

pca_cca.go
```diff
@@ -0,0 +1,239 @@
...
+
+	x, y, c *mat64.SVD
+}
+
+// Corrs returns the canonical correlations in dest.
+func (c CC) Corrs(dest []float64) []float64 {
+	return c.c.Values(dest)
+}
+
+// VectorsX returns the left-eigenvectors of the canonical correlation
```
kortschak: Corrs and Vectors* panic if CC is an unsuccessful result. This is OK since there was an error, but the panics come from mat64 which is probably not as helpful as could be. Can you check that the relevant fields are nil and provide a more informative panic. Also add a note to CC that it must not be used unless it's the result of a successful call to CanonicalCorrelations.

pca_cca.go
Armadilloa16: Okie Dokie, it kinda irks me that in the comments things need to be defined before they are used, and in the code its the other way around. Feels like it should be a consistent style thing.

Armadilloa16:
`VectorsLeft` and `VectorsRight` or just `Left` and `Right`?

pca_cca.go
Armadilloa16: but `x`, `y`, and `c` don't? None of these are outwards facing. 
`n` is the number of observations in the data, so the number of 
rows in x and/or y. `nObs` maybe?

pca_cca.go
Armadilloa16: I don't follow?

pca_cca.go
kortschak:
```
// n is the number of observations used
// to construct the canonical correlations.
n int
```

pca_cca.go
kortschak: Sorry. I'm wrong.

pca_cca.go
```diff
@@ -116,7 +106,10 @@ func (c CC) VectorsX(spheredSpace bool) (vecs *mat64.Dense) {
 // VectorsX returns the right-eigenvectors of the canonical correlation
```
kortschak: Update doc comments.

pca_cca.go
```diff
@@ -61,25 +61,8 @@ func PrincipalComponents(a mat64.Matrix, weights []float64) (vecs *mat64.Dense,
...
-	col := make([]float64, n)
-	for j := 0; j < d; j++ {
-		mat64.Col(col, j, cols)
-		floats.Scale(math.Sqrt(1/vals[j]), col)
-		cols.SetCol(j, col)
-	}
-}
-
-// CC holds the relevant SVDs for canonical correlation analysis
+// CC holds the relevant SVDs from canonical correlation analysis.
```
kortschak:
```
// CC is the result of a canonical correlation analysis. A CC should not be used unless it is the result
// of a successful call to CanonicalCorrelations.
```

pca_cca.go
```diff
@@ -0,0 +1,239 @@
...
+// Corrs returns the canonical correlations in dest.
+func (c CC) Corrs(dest []float64) []float64 {
+	return c.c.Values(dest)
+}
+
+// VectorsX returns the left-eigenvectors of the canonical correlation
+// matrix if spheredSpace is true, if spheredSpace is false it returns
+// these eigenvectors back-transformed to the original data space.
+func (c CC) VectorsX(spheredSpace bool) (vecs *mat64.Dense) {
+
```
Armadilloa16: I don't know, I'm not sure it adds much but it does seem to be the convention?

kortschak: Where?

pca_cca.go
Armadilloa16: I don't know, I'm not sure it adds much but it does seem to be the convention?

pca_cca.go
```diff
@@ -61,9 +61,11 @@ func PrincipalComponents(a mat64.Matrix, weights []float64) (vecs *mat64.Dense,
 	return vecs, vars, true
 }
 
-// CC holds the relevant SVDs from canonical correlation analysis.
-// Should not be used in the case of an unsuccessful call to CanonicalCorrelations.
+// CC is the result of a canonical correlation analysis. A CC should not be used
+// unless it is of a successful call to CanonicalCorrelations.
```
kortschak: `// unless it has been returned by a successful call to CanonicalCorrelations.`

pca_cca.go
```diff
@@ -0,0 +1,252 @@
...
+	yv.VFromSVD(c.y)
+
+	scaleColsReciSqrt(&yv, ys)
+
+	qv.Product(&yv, yv.T(), &qv)
+	qv.Scale(math.Sqrt(float64(c.n-1)), &qv)
+	return &qv
+}
+
+// CanonicalCorrelations returns c which can provide the results of canonical
```
kortschak: ...returns a CC which can...

pca_cca.go
```diff
@@ -0,0 +1,252 @@
...
+// correlation analysis. The canonical correlation matrix is calculated
+// using the singular value decompositions of the centered input.
+//
+// CanonicalCorrelations will panic if the inputs x and y do not have the same
```
kortschak: This should probably be two sentences.

pca_cca.go
```diff
@@ -0,0 +1,252 @@
...
+// number of rows, and their columns should be interpretable as two different
+// sets of measurements on the same observations.
+//
+// The slice weights is used to weight the observations. If weights is nil,
+// each weight is considered to have a value of one, otherwise the length of
+// weights must match the number of observations or CanonicalCorrelations will
+// panic.
+//
+// Xc and Yc denote the centered input data matrices x and y (column means subtracted
```
kortschak: This starts too abruptly; it reads as though Xc and Yc are function parameters which they are not. A lead in saying that this is an explanation of the process would help.

pca_cca.go
```diff
@@ -0,0 +1,252 @@
...
+import (
+	"errors"
+	"math"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix"
+	"github.com/gonum/matrix/mat64"
+)
+
+// PrincipalComponents returns the principal component direction vectors and
```
vladimir-ch: How about replacing this with
```
// PrincipalComponents returns the principal component vectors and the variances
// of the principal component scores of the input data which is represented as
// an n×d matrix a where each row is an observation and each column is a variable.
//
// The slice weights is used to weight the observations. If weights is nil, each
// weight is considered to have a value of one, otherwise the length of weights
// must match the number of observations or PrincipalComponents will panic.
//
// On successful completion, the principal component vectors are returned in
// the columns of vecs as a d×min(n,d) matrix, and the column variances of
// the principal component scores, vecs * a^T, are returned in vars as
// a min(n,d)-long slice in descending sort order.
// 
// On failure, vecs and vars are returned nil and ok is returned false.
```
I omitted the reference to SVD and the remark about centering and not scaling. Feel free to put it back. 

I also fixed the missing transpose in `vecs * a^T` (please check that I'm not wrong).

Armadilloa16:
Thanks for the improvements! I've incorporated them all and also to satisfy my OCD formatted all the comments to be 80 characters long.

pca_cca.go
Armadilloa16: Yep looks good, I left the comment
`// PrincipalComponents centers the variables but does not scale the variance.`
though as this seems relevant.

cca_test.go
```diff
@@ -0,0 +1,171 @@
...
+				-0.1666196759760772, -0.3637393866139658,
+				-0.0915512109649727, 0.1077863777929168,
+			}),
+			epsilon: 1e-12,
+		},
+		// Test results compared to those results presented in examples by
+		// Koch, Inge. Analysis of multivariate and high-dimensional data.
+		// Vol. 32. Cambridge University Press, 2013. ISBN: 9780521887939
+		{ // Boston Housing Data of Harrison and Rubinfeld (1978)
+			xdata:     bostonData.View(0, 0, 506, 7), // Per capita crime rate by town, Proportion of non-retail business acres per town, Nitric oxide concentration (parts per 10 million), Weighted distances to Boston employment centres, Index of accessibility to radial highways, Pupil-teacher ratio by town, Proportion of blacks by town
```
kortschak: Please split these two comments over multiple lines and place them above the field declarations.

pca_cca.go
```diff
@@ -0,0 +1,253 @@
...
+	for i, v := range vars {
+		vars[i] = f * v * v
+	}
+	return vecs, vars, true
+}
+
+// CC is the result of a canonical correlation analysis. A CC should not be used
+// unless it has been returned by a successful call to CanonicalCorrelations.
+type CC struct {
+	// n is the number of observations used to construct the canonical
```
kortschak: Please make this:

```
// n is the number of observations used to
// construct the canonical correlations.
```

pca_cca.go
```diff
@@ -0,0 +1,253 @@
...
+	n, _ := x.Dims()
+	yn, _ := y.Dims()
+	if n != yn {
+		panic("stat: unequal number of observations")
+	}
+	if weights != nil && len(weights) != n {
+		panic("stat: len(weights) != observations")
+	}
+
+	// Centre and factorize x and y.
```
kortschak: s/Centre/Center/ (sigh)

boston_data_test.go
```diff
@@ -0,0 +1,531 @@
...
+import "github.com/gonum/matrix/mat64"
+
+// Boston Housing Data of Harrison and Rubinfeld (1978)
+// http://dx.doi.org/10.1016/0095-0696(78)90006-2
+// http://lib.stat.cmu.edu/datasets/boston
+// Columns are;
+//  per capita crime rate by town,
+//  proportion of non-retail business acres per town,
+//  nitric oxide concentration (parts per 10 million),
+//  weighted distances to Boston employment centres,
```
kortschak: Is the spelling of centre here from the paper? If not, please use American spelling.

pca_cca.go
```diff
@@ -0,0 +1,253 @@
...
+//
+// CanonicalCorrelations will panic if the inputs x and y do not have the same
+// number of rows. Columns of x and y should be interpretable as two different
+// sets of measurements on the same observations.
+//
+// The slice weights is used to weight the observations. If weights is nil, each
+// weight is considered to have a value of one, otherwise the length of weights
+// must match the number of observations or CanonicalCorrelations will panic.
+//
+// Before describing canonical correlation analysis we first introduce some
```
kortschak: I'm still not wildly happy with the text from here on down. What I would like to see is first a broad description of what the method does, followed by the maths in a separate para, followed by how the accessors allow the user to gain the results. This should all probably be after the first paragraph above.

Try to avoid using personal pronouns - the sentence here is well suited to a maths paper, but does not fit the style for our docs.

kortschak: Something like this (which is incomplete and still not ideal):

```
// CanonicalCorrelations returns a CC which can provide the results of canonical
// correlation analysis. The canonical correlation matrix is calculated using
// the singular value decompositions of the centered input.
//
// Canonical correlation analysis finds linear combinations of pairs of random
// variables in a pair of joint dataset which have maximum correlation with each
// other. {{Explanation of sphering}}.
//
// When Xc and Yc denote the centered input data matrices x and y, Sx and Sy
// denote the sample covariance matrices within x and y respectively, and Sxy
// denotes the covariance matrix between x and y. The sphered data are
// Xc * Sx^(-1/2) and Yc * Sy^(-1/2) respectively, and the correlation matrix
// between the sphered data is the canonical correlation matrix,
// Sx^(-1/2) * Sxy * Sy^(-1/2). In cases where S^(-1/2) is ambiguous for some
// covariance matrix S, S^(-1/2) is taken to be E * D^(-1/2) * E' where S can
// be eigendecomposed as S = E * D * E'.
//
// The canonical correlations are the correlations between the corresponding
// pairs of canonical variables and can be obtained with c.Corrs(). Canonical
// variables can be obtained by projecting the sphered data into the left and
// right eigenvectors of the canonical correlation matrix, and these
// eigenvectors can be obtained with c.VectorsX(true) and c.VectorsY(true)
// respectively.
//
// The first pair of left and right eigenvectors of the canonical correlation
// matrix can be interpreted as directions into which the respective sphered
// data can be projected such that the correlation between the two projections
// is maximised. The second pair and onwards solve the same optimization but
// under the constraint that they are uncorrelated to previous projections.
//
// CanonicalCorrelations will panic if the inputs x and y do not have the same
// number of rows. Columns of x and y should be interpretable as two different
// sets of measurements on the same observations.
//
// The slice weights is used to weight the observations. If weights is nil, each
// weight is considered to have a value of one, otherwise the length of weights
// must match the number of observations or CanonicalCorrelations will panic.
//
// More details can be found at
// https://en.wikipedia.org/wiki/Canonical_correlation
// or in Chapter 3 of
// Koch, Inge. Analysis of multivariate and high-dimensional data.
// Vol. 32. Cambridge University Press, 2013. ISBN: 9780521887939
```

cca_test.go
Armadilloa16: Do you mean move all the comments above the for loop?

boston_data_test.go
Armadilloa16: fixed. The original spelling from the paper was american (it was from Boston), but the spelling in the textbook I was working off was australian (published in england). Blah.

pca_cca.go
Armadilloa16: Ok I had a go at a first pass, re-organising as you suggested and writing a terse introductory second paragraph. PTAL? 

Also, should I use A' or A^T for transpose, as this is currently inconsistent I believe.

cca_test.go
kortschak: No. Make it like so:

```
// Per capita crime rate by town, Proportion of non-retail business acres per town,
// Nitric oxide concentration (parts per 10 million), Weighted distances to Boston employment
// centers, Index of accessibility to radial highways, Pupil-teacher ratio by town,
// Proportion of blacks by town
xdata:     bostonData.View(0, 0, 506, 7),
```

with appropriate line breaks. (Also the next line like this).

cca_test.go
```diff
@@ -100,8 +102,18 @@ func TestCanonicalCorrelations(t *testing.T) {
 		// Koch, Inge. Analysis of multivariate and high-dimensional data.
 		// Vol. 32. Cambridge University Press, 2013. ISBN: 9780521887939
 		{ // Boston Housing Data of Harrison and Rubinfeld (1978)
-			xdata:     bostonData.View(0, 0, 506, 7), // Per capita crime rate by town, Proportion of non-retail business acres per town, Nitric oxide concentration (parts per 10 million), Weighted distances to Boston employment centres, Index of accessibility to radial highways, Pupil-teacher ratio by town, Proportion of blacks by town
-			ydata:     bostonData.View(0, 7, 506, 4), // Average number of rooms per dwelling, Proportion of owner-occupied units built prior to 1940, Full-value property-tax rate per $10000, Median value of owner-occupied homes in $1000s
+			xdata: bostonData.View(0, 0, 506, 7),
+			// Per capita crime rate by town,
```
kortschak: Comments above the value they are describing.

pca_cca.go
```diff
@@ -133,21 +133,18 @@ func (c CC) Right(spheredSpace bool) (vecs *mat64.Dense) {
...
-// number of rows. Columns of x and y should be interpretable as two different
-// sets of measurements on the same observations.
-//
-// The slice weights is used to weight the observations. If weights is nil, each
-// weight is considered to have a value of one, otherwise the length of weights
-// must match the number of observations or CanonicalCorrelations will panic.
+// Canonical correlation analysis can be thought of an approach to finding
+// associations between two sets of variables on the same observations. First,
+// correlations between variables within each set are removed by sphereing both
+// datasets. Second, the linear combinations of the two sets of variables that
```
kortschak: s/. Second,/, then/

pca_cca.go
```diff
@@ -133,21 +133,18 @@ func (c CC) Right(spheredSpace bool) (vecs *mat64.Dense) {
...
-// CanonicalCorrelations will panic if the inputs x and y do not have the same
-// number of rows. Columns of x and y should be interpretable as two different
-// sets of measurements on the same observations.
-//
-// The slice weights is used to weight the observations. If weights is nil, each
-// weight is considered to have a value of one, otherwise the length of weights
-// must match the number of observations or CanonicalCorrelations will panic.
+// Canonical correlation analysis can be thought of an approach to finding
+// associations between two sets of variables on the same observations. First,
+// correlations between variables within each set are removed by sphereing both
```
kortschak: s/removed by sphereing both datasets/remove in a process called sphering/ (spelling?)

pca_cca.go
```diff
@@ -156,14 +153,25 @@ func (c CC) Right(spheredSpace bool) (vecs *mat64.Dense) {
 // pairs of canonical variables and can be obtained with c.Corrs(). Canonical
 // variables can be obtained by projecting the sphered data into the left and
 // right eigenvectors of the canonical correlation matrix, and these
-// eigenvectors can be obtained with c.VectorsX(true) and c.VectorsY(true)
-// respectively.
+// eigenvectors can be obtained with c.Left(true) and c.Right(true)
+// respectively. The canoncial variables can also be obtained directly from the
+// raw (centered) data by using the back-transformed eigenvectors which can be
```
kortschak: s/raw (centered)/centered raw/

pca_cca.go
```diff
@@ -156,14 +153,25 @@ func (c CC) Right(spheredSpace bool) (vecs *mat64.Dense) {
 // pairs of canonical variables and can be obtained with c.Corrs(). Canonical
 // variables can be obtained by projecting the sphered data into the left and
 // right eigenvectors of the canonical correlation matrix, and these
-// eigenvectors can be obtained with c.VectorsX(true) and c.VectorsY(true)
-// respectively.
+// eigenvectors can be obtained with c.Left(true) and c.Right(true)
+// respectively. The canoncial variables can also be obtained directly from the
```
kortschak: s/canoncial/canonical/

Armadilloa16:
Like this?

kortschak:
Yeah, except gofmt'd :)

pca_cca.go
```diff
@@ -0,0 +1,262 @@
...
+	xsvd, ok := svdFactorizeCentered(x, weights)
+	if !ok {
+		return CC{}, errors.New("stat: failed to factorize x")
+	}
+	ysvd, ok := svdFactorizeCentered(y, weights)
+	if !ok {
+		return CC{}, errors.New("stat: failed to factorize y")
+	}
+	var xu, xv, yu, yv mat64.Dense
+	xu.UFromSVD(xsvd)
```
vladimir-ch: Do we need to take care of the case when `n < d` as in `PrincipalComponents`?

pca_cca.go
```diff
@@ -0,0 +1,262 @@
...
+	scaleColsReciSqrt(&yv, ys)
+
+	qv.Product(&yv, yv.T(), &qv)
+	qv.Scale(math.Sqrt(float64(c.n-1)), &qv)
+	return &qv
+}
+
+// CanonicalCorrelations returns a CC which can provide the results of canonical
+// correlation analysis. The canonical correlation matrix is calculated using
+// the singular value decompositions of the centered input.
```
vladimir-ch: Just a couple of quick thoughts, I'd like to come back to this later (I'm a binge reviewer these days).

How about
```
// CanonicalCorrelations returns a CC which can provide the results of canonical
// correlation analysis of the input data x and y, optionally weighted by weights.
// The canonical correlation matrix is calculated using
// the singular value decompositions of the centered input.
```
I'm not sure whether mentioning SVD is not an implementation detail and also whether the word "centered" does not suggest that the input should already be centered. Perhaps "centered" can be omitted and the rest, if kept, moved towards the end of the comment.

pca_cca.go
```diff
@@ -0,0 +1,262 @@
...
+//
+// Canonical correlation analysis can be thought of an approach to finding
```
vladimir-ch: `... analysis is an approach ...`?

pca_cca.go
```diff
@@ -0,0 +1,262 @@
...
+// associations between two sets of variables on the same observations. First,
```
vladimir-ch: This describes CCA as an algorithm, first this, then that. Isn't it better to describe here what the problem of CCA consists of? Something like
```
The goal of CCA is to find a transformation of each set of variables,
called sphered data, that maximises their correlation under the
constraint that the variables within each set of sphered data are uncorrelated.
```
I'm not sure if this is better or even a good description. Take it just as a suggestion, this topic is out of my field.

pca_cca.go
```diff
@@ -0,0 +1,262 @@
...
+// correlations between variables within each set are removed by a 
+// transformation that produces what is called the sphered data in each case,
+// then the linear combinations of the two sets of sphered variables that are 
+// maximally correlated to each other are found.
+//
+// Some notation: let Xc and Yc denote the centered input data matrices x
+// and y (column means subtracted from each column), let Sx and Sy denote the
+// sample covariance matrices within x and y respectively, and let Sxy denote
+// the covariance matrix between x and y. The sphered data can then be expressed
+// as Xc * Sx^(-1/2) and Yc * Sy^(-1/2) respectively, and the correlation matrix
```
vladimir-ch: So basically this gives the solution to the CCA problem I tried to state above.  `Sx^{-1/2}` seems to be the transformation of X (well, Xc), and similarly for Y. If we want to go into the trouble of writing it up, it would be maybe elegant to write first a statement of the CCA problem (a constrained maximization problem like what I wrote above) and this paragraph could describe its solution because it can be found explicitly (as described here).

As a side note, we don't have a consistent notation for matrix transpose, recently I use `^T` in lapack and nobody objects. But I think that we use consistently `{}` instead of `()` in cases like `D^{-1/2}`.

pca_cca.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	svd = &mat64.SVD{}
+	ok = svd.Factorize(centered, svdKind(n, d))
+	if !ok {
+		svd = nil
+	}
+	return svd, ok
+}
+
+// scaleColsReciSqrt is a helper for scaling the columns of cols by the
```
vladimir-ch: `// scaleColsReciSqrt scales the columns ...`

pca_cca.go
```diff
@@ -0,0 +1,262 @@
...
+	xv.VFromSVD(c.x)
+
+	scaleColsReciSqrt(&xv, xs)
+
+	pv.Product(&xv, xv.T(), &pv)
+	pv.Scale(math.Sqrt(float64(c.n-1)), &pv)
+	return &pv
+}
+
+// Right returns the right-eigenvectors of the canonical correlation matrix if
```
vladimir-ch: Without `-` and split into two sentences?

pca_cca.go
```diff
@@ -0,0 +1,262 @@
...
+// Corrs returns the canonical correlations in dest.
+func (c CC) Corrs(dest []float64) []float64 {
+	if c.c == nil {
+		panic("stat: canonical correlations missing or invalid")
+	}
+
+	return c.c.Values(dest)
+}
+
+// Left returns the left-eigenvectors of the canonical correlation matrix if
```
vladimir-ch: Without `-` and split into two sentences?

pca_cca.go
Armadilloa16: Good point!

This also reminds me of something I noticed in PrincipalComponents but forgot to fix -- `n < d`
should actually be `n - 1 < d` and everwhere that `min(n,d)` is used should be `min(n-1,d)` because the columns are centred this affects the upper bound on the rank of the matrix. I'll make this fix. 

While we're talking about rank -- although this is a correct upper bound it is not neccessarily correct, and I believe the usual way of estimating the rank of a matrix is using an epsilon around zero to could how many non-zero singular values it has. Is this something worth considering/ that would be useful?

Armadilloa16: Or actually... should this fix go in a seperate PR?

pca_cca.go
Armadilloa16: Yep mention of SVD doesnt need to be mentioned at all, and centering is covered under sphereing later and so is not needed here.

pca_cca.go
Armadilloa16: This is more or less what I was aiming for -- first a broad description, then the details. I tried re-writing the description part, PTAL?

pca_cca.go
Armadilloa16: So after some thought, this is not actually needed here because these vectors are not returned -- they are used to calculate `ccor`. It is the SVD objects that are returned, so no need to filter here for singular values of zero here. But in `PrincipleComponents` the changes I mentioned above should still be made.

pca_cca.go
```diff
@@ -0,0 +1,259 @@
...
+//
+// PrincipalComponents centers the variables but does not scale the variance.
+//
+// The slice weights is used to weight the observations. If weights is nil, each
+// weight is considered to have a value of one, otherwise the length of weights
+// must match the number of observations or PrincipalComponents will panic.
+//
+// On successful completion, the principal component vectors are returned in the
+// columns of vecs as a d×min(n,d) matrix, and the column variances of the
+// principal component scores, vecs * a^T, are returned in vars as a
```
vladimir-ch: I think I made you change this once already but looking at it again now it seems to me that `a * vecs` should be correct. But please better check it.

cca_example_test.go
```diff
@@ -0,0 +1,164 @@
...
+	// appropriate for canonical correlation analysis. The columns (variables)
+	// of xdata are, in order:
+	//  per capita crime rate by town,
+	//  proportion of non-retail business acres per town,
+	//  nitric oxide concentration (parts per 10 million),
+	//  weighted distances to Boston employment centres,
+	//  index of accessibility to radial highways,
+	//  pupil-teacher ratio by town, and
+	//  proportion of blacks by town.
+	xdata := bostonData.View(0, 0, n, xd)
```
kortschak: Note gonum/matrix#416. This should be rewritten as `xdata := bostonData.Slice(0, n, 0, xd)`.

cca_example_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+	// The columns (variables) of ydata are, in order:
+	//  average number of rooms per dwelling,
+	//  proportion of owner-occupied units built prior to 1940,
+	//  full-value property-tax rate per $10000, and
+	//  median value of owner-occupied homes in $1000s.
+	ydata := bostonData.View(0, xd, n, yd)
```
kortschak: `ydata := bostonData.Slice(0, n, xd, xd+yd)`

cca_test.go
```diff
@@ -0,0 +1,183 @@
...
+				12.7209032660734298, -7.6793888180353775,
+			}),
+			epsilon: 1e-12,
+		},
+		// Test results compared to those results presented in examples by
+		// Koch, Inge. Analysis of multivariate and high-dimensional data.
+		// Vol. 32. Cambridge University Press, 2013. ISBN: 9780521887939
+		{ // ASA Car Exposition Data of Ramos and Donoho (1983)
+			// Displacement, Horsepower, Weight
+			xdata: carData.View(0, 0, 392, 3),
```
kortschak: `carData.Slice(0, 392, 0, 3)`

cca_test.go
```diff
@@ -0,0 +1,183 @@
...
+			// Acceleration, MPG
+			ydata:     carData.View(0, 3, 392, 2),
```
kortschak: `carData.Slice(0, 392, 3, 5`)

cca_test.go
```diff
@@ -0,0 +1,183 @@
...
+		// Koch, Inge. Analysis of multivariate and high-dimensional data.
+		// Vol. 32. Cambridge University Press, 2013. ISBN: 9780521887939
+		{ // Boston Housing Data of Harrison and Rubinfeld (1978)
+			// Per capita crime rate by town,
+			// Proportion of non-retail business acres per town,
+			// Nitric oxide concentration (parts per 10 million),
+			// Weighted distances to Boston employment centres,
+			// Index of accessibility to radial highways,
+			// Pupil-teacher ratio by town, Proportion of blacks by town
+			xdata: bostonData.View(0, 0, 506, 7),
```
kortschak: `bostonData.View(0, 506, 0, 7)`

cca_test.go
```diff
@@ -0,0 +1,183 @@
...
+			// Average number of rooms per dwelling,
+			// Proportion of owner-occupied units built prior to 1940,
+			// Full-value property-tax rate per $10000,
+			// Median value of owner-occupied homes in $1000s
+			ydata:     bostonData.View(0, 7, 506, 4),
```
kortschak: `bostonData.View(0, 506, 7, 11)`

cca_example_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"fmt"
+	"log"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix/mat64"
+	"github.com/gonum/stat"
+)
+
+// symView is a helper for getting a View of a SymDense.
+type symView struct {
```
kortschak: There is now mat64.SymDense.SquareSlice so this can go away.

pca_cca.go
Armadilloa16: You are right. But also, `a` shouldn't be a, but rather the centered version of `a`. I'll replace
`vecs * a^T` with `b * vecs where b is a with centered columns`.

cca_example_test.go
Armadilloa16: Yep this also applies to PrincipalComponents so I'll change it there too.

cca_example_test.go
Armadilloa16: mat64.SymDense.SquareSlice doesn't do what I need here -- which is to take a non-symmetric submatrix of a SymDense. mat64.SymDense.SquareSlice only allows for symmetric submatrices (along the diagonal). Or did I misunderstand?

kortschak: No, you're right. Sorry.

Armadilloa16:
No, thank you! I learnt alot. Should I squash the commits together then?

btracey:
Sorry for being very late to this review, and if this has already been covered.

1) The types in `mat64` have their creation as a method, not a function. It's `chol.Factorize`, not `FactorizeChol`. Here, CC gets a production function. Is there a good reason for the API styles to be different? Should this change, or mat64?

2) Is there a good reason PCA isn't also a type like CC? PCA also has extra computations that may not be necessary for all users (someone who just wants the values but not the vectors)?

kortschak:
1. The types in mat64 do this for reuse and allocation savings. I guess it depends on how frequently this gets performed in loops, but there is no real reason the APIs have to look the same in that way otherwise.

2. That could be changed, but this PR has lived a long a fruitful life, perhaps that could happen in another.

btracey:
SGTM

kortschak:
OK, @Armadilloa16 can you squash and rebase this onto master.

@btracey I had a look at how we do things internally, and I doubt there is much saving from reuse because of the SVD which does allocations without considering the receiver's state - though there is a TODO from you that might change this. If you think that will change, then we should consider the making these methods on a type - in that case would you file an issue to change these. Otherwise well leave them as they are.

I'll file an issue to change PCA to only return the vectors if asked.

Armadilloa16:
Cool, and yeah for the record I agree there are improvements that could be made both in consistency of style and efficiency...

Issue #126, opened 2016-10-03
stat: Add PanicTests for stat.go file


Dawny33:
Panic tests are needed for the `stat.go` file, which are not included currently: https://coveralls.io/builds/8112223/source?filename=stat.go

The `LinearRegression` and related modules do not have panic tests. So, including them would help cover most of the file properly.

kortschak:
I think this is unnecessary.

PR #127, opened 2016-10-03
Added Panic tests for Regression Implementations

Dawny33:
Addresses https://github.com/gonum/stat/issues/126


stat_test.go
```diff
@@ -90,6 +90,64 @@ func TestCorrelation(t *testing.T) {
...
+
+		x:       make([]float64, 3),
+		y:       make([]float64, 2),
+		weights: nil,
+	},
+}
+
+func TestLinearRegressionLengthMatch(t *testing.T) {
+	for _, test := range linearRegressionMismatchTests {
+		if !Panics(func() { LinearRegression(test.x, test.y, test.weights, true) }) {
```
kortschak: Why do each of these appear twice?

Dawny33: @kortschak Working on it. Removing the redundant repetition.

Dawny33:
@kortschak I have fixed the redundant repetitions. 

But, what do you mean by:
>it would be more useful to add a test that exercised the weights branch in LinearRegression and RNoughtSquared

Both of their tests do take a legit `weights` slice into account.

kortschak:
There is testing for weighted regression when origin is false (see [here](https://coveralls.io/builds/8112223/source?filename=stat.go#L676)).

Dawny33:
@kortschak Sorry, I did not understand what you meant by it. :) Do you mean that there is already a test for weighted regression when the origin is false?

If yes, then that does not test the length mismatch panic of `weights`.

Or are you asking me to include a test for the #L676 too?

Dawny33:
@kortschak Nevermind my previous comment. I have written tests for covering the weights assumption in both `LinearRegression` and `RNoughtSquared` modules. 

Pl review.

kortschak:
I still don't really understand why you are duplicating the linear regression test function that already exists, particularly for the weighting tests. Please just add a panics girls to the test struct that currently exists and use the TestLinearRegression function with modification.

Dawny33:
@kortschak Makes sense :)

I've now removed the extra function, and added the panics to the existing struct.

Did the tests in the `TestLinearRegression` func using if conditions on the `name` field of the struct

stat_test.go
```diff
@@ -836,33 +837,89 @@ var linearRegressionTests = []struct {
...
+	},
 }
 
 func TestLinearRegression(t *testing.T) {
 	for _, test := range linearRegressionTests {
-		alpha, beta := LinearRegression(test.x, test.y, test.weights, test.origin)
-		var r float64
-		if test.origin {
-			r = RNoughtSquared(test.x, test.y, test.weights, beta)
+		if strings.Contains(test.name, "mismatch") {
```
btracey: Rather than do this strings.Contains, just add a `panics bool` field to linearRegressionTest.

Then you can do
var alpha, beta float64
if !Panics(stuff) && test.panics {
     t.Errorf("LinearRegression unexpected success")
    continue
}
````
And then the tests that are there now.

Comment (Dawny33) [inline stat_test.go]: @btracey I think it'd be more cumbersome, cause adding panics field for all of them, and then again running multiple tests inside the block with different `origin` and separate blocks for `LinearRegression`, `RSquared` and `RNoughtSquared`.

The current implementation just does a simple string match and goes ahead with the tests, which is a better implementation IMHO. What do you think?

Comment (btracey) [inline stat_test.go]: I'm not sure I understand. Only one panic field is needed, since they either all panic or they don't. It's just as easy to add a panic field, and it's much clearer. It seems like that field could be a direct replacement for the Contains call.

File: stat_test.go [hunk 3]
```diff
@@ -836,33 +837,89 @@ var linearRegressionTests = []struct {
...
+			}
+			if !Panics(func() { RSquared(test.x, test.y, test.weights, 0, 0) }) {
+				t.Errorf("RSquared did not panic with lengths length mismatch")
+			}
+			if !Panics(func() { RSquaredFrom(test.x, test.y, test.weights) }) {
+				t.Errorf("RSquaredFrom did not panic with weights length mismatch")
+			}
+			if !Panics(func() { RNoughtSquared(test.x, test.y, test.weights, 0) }) {
+				t.Errorf("RNoughtSquared did not panic with weights length mismatch")
+			}
```
btracey: After the panics tests, put a continue, then you can remove the else.

stat_test.go
```diff
@@ -836,33 +837,89 @@ var linearRegressionTests = []struct {
...
+		} else if test.name == "weights are nil" {
```
btracey: Remove this else if. As above, I don't like the string match, and hand-coding the weight size is not good practice. The test isn't bad, but it should be put after the checks below. At that point you can do

if test.weights != nil { 
    continue
}
weights = make([]float64, len(test.x))
 for i := range weights {
      weights[i] = 1
}
````
Then make the calls and check that the sizes match.

btracey:
Ping?

btracey:
Closing due to inactivity.

PR #130, opened 2016-10-15
Add Students T distribution

btracey:


kortschak:
Where are these random regressions coming from?

btracey:
I don't know. I'm looking into it now. It seems something like sampling, but they all have a fixed seed...

btracey:
It's partly the change to the call to quad.Fixed, but that's not a sufficient explanation.

btracey:
I'm going to submit these two PRs and then figure out what's going on with the tests

PR #132, opened 2016-10-18
Improve implementation of Gamma random number generation

btracey:


btracey:
The tests all pass on my machine at least, we'll see about travis.

These algorithm choices are better than the previous ones. In particular, for alpha > 1, the old algorithm was implemented incorrectly, and required generating ~alpha random numbers per trial. This is much better.

The alpha < 1 case in split into two cases. alpha < 0.3 has bad behavior near alpha = 1, and the other case has bad behavior for alpha small. 

I'm not sure how to test the alpha < 0.3 case. The problem is that the PDF is poorly behaved, and so the quadrature is very inaccurate. I confirmed that it works for the alpha cases  here.

btracey:
Also, if you see the problem with Algorithm 3 I'll happily change it, but I don't see what's wrong.

distuv/gamma.go
```diff
@@ -95,56 +95,112 @@ func (g Gamma) Rand() float64 {
...
+		//  Kundu, Debasis, and Rameshwar D. Gupta. "A convenient way of generating
+		//  gamma random variables using generalized exponential distribution."
+		//  Computational Statistics & Data Analysis 51.6 (2007): 2796-2802.
+
+		// TODO(btracey): Change to using Algorithm 3 if we can find the bug in
+		// the implementation below.
+
+		// Algorithm 2.
+		alpha := g.Alpha
+		a := math.Pow(1-math.Exp(-1.0/2), alpha) / (math.Pow(1-math.Exp(-1.0/2), alpha) + alpha*math.Exp(-1)/math.Pow(2, alpha))
```
kortschak: Have predefined vars for the `math.Exp(const)` values here?

distuv/gamma.go
```diff
@@ -95,56 +95,112 @@ func (g Gamma) Rand() float64 {
...
+				if v <= math.Pow(x, alpha-1) {
+					break
+				}
+			}
 		}
-		return t / a / b
+		return x / g.Beta
+
+		/*
+			//  Algorithm 3.
```
kortschak: I can't see any error here compared to the paper.

distuv/gamma.go
```diff
@@ -95,56 +95,112 @@ func (g Gamma) Rand() float64 {
...
-		} else {
-			bp = b * (ap - 1) / (a - 1)
-			lkp = (ap - a) + (a-ap)*(a-1)/b
-		}
+		// Generate using:
+		//  Marsaglia, George, and Wai Wan Tsang. "A simple method for generating
+		//  gamma variables." ACM Transactions on Mathematical Software (TOMS)
+		//  26.3 (2000): 363-372.
+		d := a - 1.0/3
+		sqrt9d := 3 * math.Sqrt(d)
```
kortschak: Make this the reciprocal since `a*b` is faster than `a/b`. `c := 1 / (3*math.Sqrt(d))` and use below. This also matches the paper better.

distuv/gamma.go
btracey: Done.

PR #133, opened 2016-10-21
Addmvt

btracey:


btracey:
The failures here are related to #132

btracey:
Should I merge master into this before review?

kortschak:
Can you rebase master rather than merge?

btracey:
Okay, I believe I did that.

btracey:
Working on failure

btracey:
There we go.

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+	once  sync.Once
+	sigma *mat64.SymDense // only stored if needed
+
+	chol       mat64.Cholesky
+	lower      mat64.TriDense
+	logSqrtDet float64
+	dim        int
+}
+
+// NewStudentsT creates a new StudentsT with the given scale, mean, and covariance
```
kortschak: Use the parameter names as they exist in the signature, or explain the relationship here between the scale, mean, and covariance and the parameters in the signature.

kortschak: Use the parameters as they exist in the signature, or explain the relationship between scale, mean, and covariance and the parameters there.

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+// parameters.
+//
+// NewStudentsT panics if len(mu) == 0, or if len(mu) != sigma.Symmetric(). If
+// the covariance matrix is not positive-definite, nil is returned and the
+// boolean is false.
```
kortschak: Return `(dist *StudentsT, ok bool)` so the documentation can refer to `ok`.

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+// ConditionStudentsT returns the Student's T distribution that is the receiver
+// conditioned on the input evidence. The returned Student's T has dimension
+// n - len(observed), where n is the dimension of the original receiver.
+// The dimension order is preserved during conditioning, so if the value
+// of dimension 1 is observed, the returned normal represents dimensions {0, 2, ...}
+// of the original Student's T distribution.
+//
+// ConditionStudentsT returns {nil, false} if there is a failure during the update.
```
kortschak: Use `(dist *StudentsT, ok bool)` and refer to the `ok`. I don't really like the `{nil, false}`; I don't think we have ever used that before. Add "If ok is returned false, dist is not usable." somewhere - probably before the explanation.

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+	unobserved = make([]int, 0, unob)
+	for i := 0; i < dim; i++ {
+		if _, ok := obMap[i]; !ok {
+			unobserved = append(unobserved, i)
+		}
+	}
+	return unobserved
+}
+
+// studentsTConditional updates a Students T distribution based on the observed samples
```
kortschak: "Student's T"

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+	"math"
+	"math/rand"
+	"sync"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix/mat64"
+	"github.com/gonum/stat/distuv"
+)
+
+// StudentT is a multi-variate StudentT distribution. It is a distribution over
```
kortschak: Second "StudentT" should be "Student's T" I think.

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+	newNu, newPhi, newK := studentsTConditional(observed, values, s.nu, s.mu, s.sigma)
+	if newPhi == nil {
+		return nil, false
+	}
+
+	return NewStudentsT(newNu, newPhi, newK, src)
+
+}
+
+// findUnob returns the unobserved variables (the complementary set to observed).
```
kortschak: Can you move this below `studentsTConditional`?

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+// findUnob panics if any value repeated in observed.
+func findUnob(observed []int, dim int) (unobserved []int) {
+	ob := len(observed)
+	unob := dim - ob
+	obMap := make(map[int]struct{})
```
kortschak: The [`golang.org/tools/contained/intsets.Sparse`](https://godoc.org/golang.org/x/tools/container/intsets#Sparse) would be a good fit here. It is faster than `map[int]struct{}` for this kind of thing.

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+			unobserved = append(unobserved, i)
+		}
+	}
+	return unobserved
+}
+
+// studentsTConditional updates a Students T distribution based on the observed samples
+// (see documentation for the public function). The Gaussian conditional update
+// is treated as a special case when  nu == math.Inf(1).
+func studentsTConditional(observed []int, values []float64, nu float64, mu []float64, sigma mat64.Symmetric) (newNu float64, newMean []float64, newSigma mat64.Symmetric) {
```
kortschak: This can return a concrete `*mat64.SymDense`, though I'm not fussed about this either way.

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+	s.ScaleSym(st.nu/(st.nu-2), s)
+	return s
+}
+
+// Dim returns the dimension of the distribution.
+func (s *StudentsT) Dim() int {
+	return s.dim
+}
+
+func (s *StudentsT) LogProb(y []float64) float64 {
```
kortschak: Doc comment.

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+	if len(y) != s.dim {
+		panic(badInputLength)
+	}
+
+	nu := s.nu
```
kortschak: I think we could have more commentary here - at least where this is from.

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+
+	return t1 - ((nu+n)/2)*math.Log(1+dot/nu)
+}
+
+// MarginalStudentsT returns the marginal distribution of the given input variables.
+// That is, MarginalStudentsT returns
+//  p(x_i) = \int_{x_o} p(x_i | x_o) p(x_o) dx_o
+// where x_i are the dimensions in the input, and x_o are the remaining dimensions.
+// The input src is passed to the call to NewStudentsT.
+func (s *StudentsT) MarginalStudentsT(vars []int, src *rand.Rand) (*StudentsT, bool) {
```
kortschak: Return `(dist *StudentsT, ok bool)`.

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+// Mean returns the mean of the probability distribution at x. If the
+// input argument is nil, a new slice will be allocated, otherwise the result
+// will be put in-place into the receiver.
+func (s *StudentsT) Mean(x []float64) []float64 {
+	x = reuseAs(x, s.dim)
+	copy(x, s.mu)
+	return x
+}
+
+func (s *StudentsT) Prob(y []float64) float64 {
```
kortschak: Doc comment.

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+	return math.Exp(s.LogProb(y))
+}
+
+func (s *StudentsT) Rand(x []float64) []float64 {
```
kortschak: Doc comment.

distmv/studentst.go
```diff
@@ -0,0 +1,335 @@
...
+	xVec.MulVec(&s.lower, tmpVec)
+
+	u := distuv.ChiSquared{K: s.nu, Src: s.src}.Rand()
+	floats.Scale(math.Sqrt(s.nu/u), x)
+
+	floats.Add(x, s.mu)
+	return x
+}
+
+// setK computes and stores the covariance matrix of the distribution.
```
kortschak: s/setK/setSigma/

btracey: Done.

distmv/studentst.go
btracey: Done.

distmv/studentst.go
btracey: Done.

distmv/studentst.go
btracey: Done.

distmv/studentst.go
btracey: Thanks.

distmv/studentst.go
btracey: Done.

distmv/studentst.go
btracey: It's in Normal, but that one can change too.

Done.

distmv/studentst.go
btracey: Done.

distmv/studentst.go
btracey: Done.

distmv/studentst.go
btracey: Done.

distmv/studentst.go
btracey: It's just the formula in the wikipedia entry, or in the pdf linked in the struct documentation. What kind of commentary would you like?

distmv/studentst.go
btracey: Done.

distmv/studentst.go
btracey: Okay, I reformulated the code using intsets.

distmv/studentst.go
```diff
@@ -0,0 +1,345 @@
...
+// conditioned on the input evidence. The returned Student's T has dimension
+// n - len(observed), where n is the dimension of the original receiver.
+// The dimension order is preserved during conditioning, so if the value
+// of dimension 1 is observed, the returned normal represents dimensions {0, 2, ...}
+// of the original Student's T distribution.
+//
+// ok indicates whether there was a failure during the update, and if ok is false
+// dist is not usable.
+// Mathematically this is impossible, but can occur with finite precision arithmetic.
+func (s *StudentsT) ConditionStudentsT(observed []int, values []float64, src *rand.Rand) (*StudentsT, bool) {
```
kortschak: dist and ok need to be named in the return for the doc to be meaningful :)

distmv/studentst.go
kortschak: I guess that's OK.

distmv/studentst.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+// findUnob returns the unobserved variables (the complementary set to observed).
+// findUnob panics if any value repeated in observed.
+func findUnob(observed []int, dim int) (unobserved []int) {
+	setOb := &intsets.Sparse{}
+	for _, v := range observed {
+		setOb.Insert(v)
+	}
+	setAll := &intsets.Sparse{}
```
kortschak: `var setAll insets.Sparse`

distmv/studentst.go
```diff
@@ -0,0 +1,345 @@
...
+	for i := 0; i < dim; i++ {
+		setAll.Insert(i)
+	}
+	setUnob := &intsets.Sparse{}
```
kortschak: `var setUnob intsets.Sparse`

distmv/studentst.go
```diff
@@ -0,0 +1,345 @@
...
+	if !ok {
+		return nil, false
+	}
+	s.lower.LFromCholesky(&s.chol)
+	s.logSqrtDet = 0.5 * s.chol.LogDet()
+	return s, true
+}
+
+// ConditionStudentsT returns the Student's T distribution that is the receiver
+// conditioned on the input evidence. The returned Student's T has dimension
```
kortschak: ...", and the success of the operation."

distmv/studentst.go
```diff
@@ -0,0 +1,345 @@
...
+// n - len(observed), where n is the dimension of the original receiver.
+// The dimension order is preserved during conditioning, so if the value
+// of dimension 1 is observed, the returned normal represents dimensions {0, 2, ...}
+// of the original Student's T distribution.
+//
+// ok indicates whether there was a failure during the update, and if ok is false
```
kortschak: "If ok is false the update failed and dist is not usable. Mathematically..."

distmv/studentst.go
```diff
@@ -0,0 +1,345 @@
...
+
+	var tmp mat64.Vector
+	tmp.SolveCholeskyVec(&s.chol, x)
+
+	dot := mat64.Dot(&tmp, x)
+
+	return t1 - ((nu+n)/2)*math.Log(1+dot/nu)
+}
+
+// MarginalStudentsT returns the marginal distribution of the given input variables.
```
kortschak: ...", and the success of the operation."

distmv/studentst.go
```diff
@@ -0,0 +1,345 @@
...
+// That is, MarginalStudentsT returns
+//  p(x_i) = \int_{x_o} p(x_i | x_o) p(x_o) dx_o
+// where x_i are the dimensions in the input, and x_o are the remaining dimensions.
+// The input src is passed to the call to NewStudentsT.
+//
+// ok indicates whether there was a failure during the update, and if ok is false
```
kortschak: "If ok is false the update failed and dist is not usable. Mathematically..."

btracey:
Maybe @jonlawlor ?

jonlawlor:
I checked the log pdf against the source paper, a python example on SO that is probably from the same paper, and it matched.  Also checks out against the R implementation.

I tested a few edge cases as well.  When it is 1 dimensional it matches the uvdist Student (I can submit the code if you want, it is simple).  If you choose nu = math.Inf, NaNs are returned instead of matching mvdist.Normal, it might be nice to special case that, because it would match the handling in the ConditionStudentsT method.  Also, for the tests is there a reason you are not using the testProbability helper function?

Otherwise I only verified that the Covariance, MarginalStudentsT, Mean, and Rand functions match the function descriptions where I could find them, usually wikipedia or SO.  No comment on the ConditionStudentsT.

Otherwise LGTM, special casing mu = math.Inf could always wait for another pull.

distmv/studentst.go
btracey: Done.

distmv/studentst.go
btracey: Thanks.

distmv/studentst.go
btracey: Such high standards! :)

distmv/studentst.go
btracey: Done. Thanks

distmv/studentst.go
btracey: Done. Thanks

distmv/studentst.go
btracey: Done.

distmv/studentst.go
btracey: Done.

distmv/studentst.go
```diff
@@ -0,0 +1,347 @@
...
+	"golang.org/x/tools/container/intsets"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix/mat64"
+	"github.com/gonum/stat/distuv"
+)
+
+// StudentsT is a multi-variate Student's T distribution. It is a distribution over
+// ℝ^n with the probability density
+//  p(y) = (Γ((ν+d)/2) / Γ(ν/2)) * (νπ)^(-n/2) * |Ʃ|^(-1/2) *
```
Armadilloa16: (v+d) should be (v + n)

Armadilloa16:
Wow this all happened in one afternoon. Should I still look through it?

kortschak:
Yes, please.

distmv/studentst.go
```diff
@@ -0,0 +1,347 @@
...
+	"sync"
+
+	"golang.org/x/tools/container/intsets"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix/mat64"
+	"github.com/gonum/stat/distuv"
+)
+
+// StudentsT is a multi-variate Student's T distribution. It is a distribution over
```
Armadilloa16: Although I actually like the term "multi-variate" as it emphasises its meaning --- multiple variables --- it is not generally hyphenated, normally it is written "multivariate".

distmv/studentst.go
```diff
@@ -0,0 +1,347 @@
...
+	mu2 := make([]float64, ob) // really v - mu2
+	for i, v := range observed {
+		mu2[i] = values[i] - mu[v]
+	}
+
+	var sigma11, sigma22 mat64.SymDense
+	sigma11.SubsetSym(sigma, unobserved)
+	sigma22.SubsetSym(sigma, observed)
+
+	sigma21 := mat64.NewDense(ob, unob, nil)
```
Armadilloa16: I needed to do something similar for the cca stuff, and I ended up making a symView helper, might be worth considering having something similar somewhere which could be used here?

distmv/studentst.go
```diff
@@ -0,0 +1,347 @@
...
+		}
+	}
+
+	var chol mat64.Cholesky
+	ok := chol.Factorize(&sigma22)
+	if !ok {
+		return math.NaN(), nil, nil
+	}
+
+	// Compute mu_1 + sigma_{2,1}^T * sigma_{2,2}^-1 (v - mu_2).
```
Armadilloa16: Is there a reason you are using sigma_{21}^T instead of just extracting sigma_{12} to begin with that I am missing?

distmv/studentst.go
```diff
@@ -0,0 +1,347 @@
...
+
+	// Compute sigma_{1,1} - sigma_{2,1}^T * sigma_{2,2}^-1 * sigma_{2,1}.
+	// TODO(btracey): Should this be a method of SymDense?
+	var tmp3, tmp4 mat64.Dense
+	err = tmp3.SolveCholesky(&chol, sigma21)
+	if err != nil {
+		return math.NaN(), nil, nil
+	}
+	tmp4.Mul(sigma21.T(), &tmp3)
+
```
Armadilloa16: The space here was abit confusing because you started the previous "paragraph" with "Compute 
sigma_{1,1} - sigma_{2,1}^T * sigma_{2,2}^-1 * sigma_{2,1}" but here you have computed 
tmp4 = sigma_{2,1}^T * sigma_{2,2}^-1 * sigma_{2,1} and then the following subtracts it from sigma_{1,1}. 
It would have been alot more clear to me if line 169 had read "Compute tmp4 which is sigma_{2,1}^T * sigma_{2,2}^-1 * sigma_{2,1}." and then here had a comment saying "compute Sigma_{1,1} - tmp4. Could be just me though.

distmv/studentst.go
btracey: Open an issue in `matrix`?

distmv/studentst.go
```diff
@@ -0,0 +1,347 @@
...
+		return math.NaN(), nil, nil
+	}
+	tmp2.MulVec(sigma21.T(), &tmp)
+
+	for i := range mu1 {
+		mu1[i] += tmp2.At(i, 0)
+	}
+
+	// Compute sigma_{1,1} - sigma_{2,1}^T * sigma_{2,2}^-1 * sigma_{2,1}.
+	// TODO(btracey): Should this be a method of SymDense?
```
btracey: Yea. I think the Schur complement is defined on the specific blocks, while this the Schur complement with respect to a set of rows. Schur complement comes up a lot (if you want to open an issue), but I'm not sure this specialized form does.

PR #135, opened 2016-10-25
Expand definition of Student's T

btracey:


btracey:
It seems reasonable to have this definition of Students T, just as we have Normal and not exclusively Standard Normal.  This also fits very well with #133, since Marginal Single  is a Students T of this form.

Issue #141, opened 2016-11-01
Any interest in implementing CircularMean?


rikonor:
I've recently had to calculate a [circular mean](https://en.wikipedia.org/wiki/Mean_of_circular_quantities) for a series of angles and noticed it's not implemented in gonum/stat.
If there's any interest I'd be happy to submit a PR for it.

jonlawlor:
I'll defer to others about where exactly it might belong, but it does appear in scipy under scipy.stats.morestats, R has it under a circular package, there's a pretty mature matlab "toolbox" on matlabcentral.  That doesn't mean we have to, but it indicates to me that it does have a certain level of utility.

It is a bit unusual (I have never used directional statistics myself) but not so specific that we would exclude it I think.  Maybe in a new sub-repository stat/circular or stat/directional ?  Thoughts anyone...

btracey:
We have a bunch of weird distances in gonum/stat. I'd just put it there unless there are a whole bunch of variations on a theme.

jonlawlor:
I did a brief survey of the methods implemented elsewhere and a few textbook table of contents and it looks like it can go from standard summary statistics (mean, var, std) on one end all the way up to recreating the statistics package using directional statistics on the other.

Maybe just put the standard summary statistics in stat and if anything goes past that put it into a sub-repository?

PR #142, opened 2016-11-01
Add CircularMean

rikonor:
In reference to #141

Apologies but I'm unsure of how to incorporate weights into the calculation.

### DIFF
--- a/stat.go
+++ b/stat.go
@@ -125,6 +125,31 @@ func ChiSquare(obs, exp []float64) float64 {
 	return result
 }
 
+// CircularMean returns the circular mean of the dataset.
+//	atan2(\sum_i w_i * sin(alpha_i), \sum_i w_i * cos(alpha_i))
+// If weights is nil then all of the weights are 1. If weights is not nil, then
+// len(x) must equal len(weights).
+func CircularMean(x, weights []float64) float64 {
+	if weights != nil && len(x) != len(weights) {
+		panic("stat: slice length mismatch")
+	}
+
+	var aX, aY float64
+	if weights != nil {
+		for i, v := range x {
+			aX += weights[i] * math.Cos(v)
+			aY += weights[i] * math.Sin(v)
+		}
+	} else {
+		for _, v := range x {
+			aX += math.Cos(v)
+			aY += math.Sin(v)
+		}
+	}
+
+	return math.Atan2(aY, aX)
+}
+
 // Correlation returns the weighted correlation between the samples of x and y
 // with the given means.
 //  sum_i {w_i (x_i - meanX) * (y_i - meanY)} / (stdX * stdY)
--- a/stat_test.go
+++ b/stat_test.go
@@ -13,6 +13,52 @@ import (
 	"github.com/gonum/floats"
 )
 
+func ExampleCircularMean() {
+	x := []float64{0, 0.25 * math.Pi, 0.75 * math.Pi}
+	weights := []float64{1, 2, 2.5}
+	cmean := CircularMean(x, weights)
+
+	fmt.Printf("The circular mean is %.5f.\n", cmean)
+	// Output:
+	// The circular mean is 1.37037.
+}
+
+func TestCircularMean(t *testing.T) {
+	for i, test := range []struct {
+		x   []float64
+		wts []float64
+		ans float64
+	}{
+		// Values compared against scipy.
+		{
+			x:   []float64{0, 2 * math.Pi},
+			ans: 0,
+		},
+		{
+			x:   []float64{0, 0.5 * math.Pi},
+			ans: 0.78539816339744,
+		},
+		{
+			x:   []float64{-1.5 * math.Pi, 0.5 * math.Pi, 2.5 * math.Pi},
+			wts: []float64{1, 2, 3},
+			ans: 0.5 * math.Pi,
+		},
+		{
+			x:   []float64{0, 0.5 * math.Pi},
+			wts: []float64{1, 2},
+			ans: 1.10714871779409,
+		},
+	} {
+		c := CircularMean(test.x, test.wts)
+		if math.Abs(c-test.ans) > 1e-14 {
+			t.Errorf("Circular mean mismatch case %d: Expected %v, Found %v", i, test.ans, c)
+		}
+	}
+	if !Panics(func() { CircularMean(make([]float64, 3), make([]float64, 2)) }) {
+		t.Errorf("CircularMean did not panic with x, wts length mismatch")
+	}
+}
+
 func ExampleCorrelation() {
 	x := []float64{8, -3, 7, 8, -4}
 	y := []float64{10, 5, 6, 3, -1}


stat.go
```diff
@@ -372,6 +372,16 @@ func HarmonicMean(x, weights []float64) float64 {
 	return math.Exp(math.Log(W) - v)
 }
 
+// CircularMean returns the circular mean of the dataset
+func CircularMean(x []float64) float64 {
```
btracey: It'll just be `aX += w[i] * math.Cos(x[i])` if the weights aren't nil. Look at Mean for an example how to code.

Please move this function to be in alphabetical order in the file.

stat_test.go
```diff
@@ -333,6 +333,32 @@ func TestHarmonicMean(t *testing.T) {
 	}
 }
 
+func ExampleCircularMean() {
+	x := []float64{.0, 2 * math.Pi, 4 * math.Pi, 6 * math.Pi, 8 * math.Pi}
```
btracey: This is a good example to have, but you should also have one where the mean is non zero.

btracey: No need for the decimal point in front of 0

stat_test.go
```diff
@@ -333,6 +333,32 @@ func TestHarmonicMean(t *testing.T) {
...
+}
+
+func TestCircularMean(t *testing.T) {
+	for i, test := range []struct {
+		x   []float64
+		ans float64
+	}{
+		{
+			x:   []float64{.0, 2 * math.Pi},
+			ans: .0,
```
btracey: No need for the decimal points in front of 0

btracey: Please also add a test with non-zero mean (and also tests with weighted data)

stat.go
```diff
@@ -372,6 +372,16 @@ func HarmonicMean(x, weights []float64) float64 {
 	return math.Exp(math.Log(W) - v)
 }
 
+// CircularMean returns the circular mean of the dataset
```
kortschak: Full stop please.

stat_test.go
```diff
@@ -333,6 +333,32 @@ func TestHarmonicMean(t *testing.T) {
 	}
 }
 
+func ExampleCircularMean() {
+	x := []float64{.0, 2 * math.Pi, 4 * math.Pi, 6 * math.Pi, 8 * math.Pi}
+	cmean := CircularMean(x)
+
+	fmt.Printf("The circular mean is %.5f.\n", cmean)
+	// Output:
+	// The circular mean is -0.00000.
```
kortschak: It's probably worth special-casing 0 so that -0 is not returned.

rikonor:
Hi all,

Thanks for the quick response - I've tried to address all the issues raised.
One question - do you all agree about the special case for 0? I've tried to implement that on [stats.go#148](https://github.com/gonum/stat/pull/142/files#diff-33838b2a8ea2f8e4311d86efea5944c5R148).

stat.go
```diff
@@ -125,6 +125,32 @@ func ChiSquare(obs, exp []float64) float64 {
 	return result
 }
 
+// CircularMean returns the circular mean of the dataset.
```
btracey: Could you provide the formula like we do in Mean? It makes it clear for those who don't know.

stat.go
```diff
@@ -125,6 +125,32 @@ func ChiSquare(obs, exp []float64) float64 {
...
+// If weights is nil then all of the weights are 1. If weights is not nil, then
+// len(x) must equal len(weights).
+func CircularMean(x, weights []float64) float64 {
+	if weights != nil && len(x) != len(weights) {
+		panic("stat: slice length mismatch")
+	}
+
+	var aX, aY float64
```
btracey: Change this as @jonlawlor  suggested, and may as well move the weights check out of the loop

````
var aX, aY float64
if weights != nil {
    for _, v := range x {
        ax += math.Cos(v)
        ay += math.Sin(v)
    }
}else{
    for i, v := range x {
        aX += weights[i]  * math.Cos(v)
        aY += weights[i] * math.Sin(v)
   }
}
````

File: stat.go [hunk 8]
```diff
@@ -125,6 +125,32 @@ func ChiSquare(obs, exp []float64) float64 {
...
+			aX += weights[i] * math.Cos(x[i])
+			aY += weights[i] * math.Sin(x[i])
+			continue
+		}
+		aX += math.Cos(x[i])
+		aY += math.Sin(x[i])
+	}
+
+	res := math.Atan2(aY, aX)
+	if math.Abs(res) < 1e-14 {
```
btracey: I'm not sure what @kortschak had in mind, but this shouldn't be based on a tolerance. Maybe

````
res := math.Atan2(aY, aX)
// correct for negative 0
if res == 0 { 
      return 0
}
````

File: stat_test.go [hunk 9]
```diff
@@ -13,6 +13,51 @@ import (
...
+		wts []float64
+		ans float64
+	}{
+		{
+			x:   []float64{0, 2 * math.Pi},
+			ans: 0,
+		},
+		{
+			x:   []float64{0, 0.5 * math.Pi},
+			ans: 0.78539816339744,
```
btracey: Are these numbers from scipy or somewhere? If so, please comment.

btracey: For example, https://github.com/gonum/stat/blob/master/distuv/studentst_test.go#L20

stat_test.go
```diff
@@ -13,6 +13,51 @@ import (
 	"github.com/gonum/floats"
 )
 
+func ExampleCircularMean() {
+	x := []float64{0, 2 * math.Pi, 4 * math.Pi, 6 * math.Pi, 8 * math.Pi}
```
btracey: Please include an example with numbers that aren't all zero. (maybe 0, pi/3, 2pi/3, pi or something)

stat.go
rikonor: Not sure I follow here - in the above snippet the `if` block seems redundant.

stat.go
rikonor: Moved the `if` out of the loop.

stat.go
rikonor: I added the formula, tried to rely on the formatting from the other formulas in `stats.go`.

stat_test.go
jonlawlor: Generally no.

rikonor:
I hope I've addressed all the issues, is there anything I missed?

stat.go
```diff
@@ -125,6 +125,35 @@ func ChiSquare(obs, exp []float64) float64 {
...
+// len(x) must equal len(weights).
+func CircularMean(x, weights []float64) float64 {
+	if weights != nil && len(x) != len(weights) {
+		panic("stat: slice length mismatch")
+	}
+
+	var aX, aY float64
+	if weights != nil {
+		for i := range x {
+			aX += weights[i] * math.Cos(x[i])
```
btracey: You still need to do two-element range syntax here (as suggested before)

stat.go
```diff
@@ -125,6 +125,35 @@ func ChiSquare(obs, exp []float64) float64 {
...
+		}
+	} else {
+		for i := range x {
+			aX += math.Cos(x[i])
+			aY += math.Sin(x[i])
+		}
+	}
+
+	res := math.Atan2(aY, aX)
+	if math.Abs(res) < 1e-14 {
```
btracey: Don't have this tolerance (as suggested before).

stat_test.go
btracey: Usually we do have a comment that we got the numbers from scipy (see the link I posted above).

stat.go
rikonor: Removed.

stat.go
rikonor: Updated. Is there a difference in performance? I've noticed other places in `stat.go` follow the one-element range syntax (which is why I followed suit).

btracey:
If you could please add the comment that the data is from scipy, then LGTM

btracey:
And if you could also add yourself to CONTRIBUTORS and/or AUTHORS though a pull request to gonum/license

stat_test.go
rikonor: Verified the results via `scipy`:
```
import math
import scipy
from scipy import stats

scipy.stats.circmean(samples=[0, math.pi / 2])
```

And added a comment explaining that.

btracey:
@jonlawlor Do you want another look?

rikonor:
Very nice. Added a PR to `gonum/license`. Thanks guys.

btracey:
Thanks for the contribution. Waiting for approval from @jonlawlor

stat.go
jonlawlor: There used to be a performance difference, but I don't know if the compiler has gotten smart enough in the last few Go versions to optimize.  Regardless of performance, [Effective Go](https://golang.org/doc/effective_go.html#for) says to use it:
> If you only need the second item in the range (the value), use the blank identifier, an underscore, to discard the first:
> ```
sum := 0
for _, value := range array {
    sum += value
}
```
So it is idiomatic and very slightly less complicated to use `_, v`.  If it occurs in other places in stat.go (or any gonum package) then that is a mistake and we should fix it.

File: stat_test.go [hunk 13]
```diff
@@ -13,6 +13,52 @@ import (
...
+	// The circular mean is 1.37037.
+}
+
+func TestCircularMean(t *testing.T) {
+	for i, test := range []struct {
+		x   []float64
+		wts []float64
+		ans float64
+	}{
+		// Valuess compared against scipy.
```
jonlawlor: s/Valuess/Values

rikonor: Oops. Fixed 👍

stat.go
rikonor: Understood. You don't see much of `x[i]` in the wild which is I why I was curious.

jonlawlor:
LGTM

PR #149, opened 2016-12-15
distmat: cleanup docs for Wishart distribution

vladimir-ch:
Some lint I missed in the review.


distmat/wishart.go
```diff
@@ -70,7 +71,8 @@ func NewWishart(v mat64.Symmetric, nu float64, src *rand.Rand) (*Wishart, bool)
 
 // MeanSym returns the mean matrix of the distribution as a symmetric matrix.
 // If x is nil, a new matrix is allocated and returned. If x is not nil, the
-// result is stored in-place into x. It must have size d×d or MeanSym will panic.
+// result is stored in-place into x. Its order must be equal to that of the
+// receiver or MeanSym will panic.
```
sbinet: `s/or/otherwise/` ?
or, better (?):
`MeanSym will panic if the order of x is not equal to the order of the receiver`

PR #151, opened 2016-12-19
stat: make CCA and PCA analysis reuse allocations

kortschak:
@btracey Please take a look.

This is a stab at a PCA API for allocation reuse. I will add CCA to this if you think this is the right way to go (and flesh out the tests and fix up docs).

One issue is the size restrictions on the input matrix for Vectors. It is not easy to be strict without being difficult because of the behaviour of mat64.SVD.

### DIFF
--- a/pca_cca.go
+++ b/pca_cca.go
@@ -13,120 +13,106 @@ import (
 	"github.com/gonum/matrix/mat64"
 )
 
-// PrincipalComponents returns the principal component vectors and the variances
-// of the principal component scores of the input data which is represented as
-// an n×d matrix a where each row is an observation and each column is a
-// variable.
+// PC is a type for computing and extracting the principal components of a
+// matrix. The results of the principal components analysis are only valid
+// if the call to PrincipalComponents was successful.
+type PC struct {
+	n, d    int
+	weights []float64
+	svd     *mat64.SVD
+	ok      bool
+}
+
+// PrincipalComponents performs a weighted principal components analysis on the
+// matrix of the input data which is represented as an n×d matrix a where each
+// row is an observation and each column is a variable.
 //
 // PrincipalComponents centers the variables but does not scale the variance.
 //
-// The slice weights is used to weight the observations. If weights is nil, each
+// The weights slice is used to weight the observations. If weights is nil, each
 // weight is considered to have a value of one, otherwise the length of weights
 // must match the number of observations or PrincipalComponents will panic.
 //
-// On successful completion, the principal component vectors are returned in the
-// columns of vecs as a d×min(n,d) matrix, and the column variances of the
-// principal component scores, b * vecs where b is a with centered columns, are
-// returned in vars as a min(n,d)-long slice in descending sort order.
-//
-// On failure, vecs and vars are returned nil and ok is returned false.
-func PrincipalComponents(a mat64.Matrix, weights []float64) (vecs *mat64.Dense, vars []float64, ok bool) {
-	n, d := a.Dims()
-	if weights != nil && len(weights) != n {
+// PrincipalComponents returns whether the analysis was successful.
+func (c *PC) PrincipalComponents(a mat64.Matrix, weights []float64) (ok bool) {
+	c.n, c.d = a.Dims()
+	if weights != nil && len(weights) != c.n {
 		panic("stat: len(weights) != observations")
 	}
 
-	svd, ok := svdFactorizeCentered(a, weights)
-	if !ok {
-		return nil, nil, false
+	c.svd, c.ok = svdFactorizeCentered(c.svd, a, weights)
+	if c.ok {
+		c.weights = append(c.weights[:0], weights...)
 	}
-
-	vecs = &mat64.Dense{}
-	vecs.VFromSVD(svd)
-	if n < d {
-		// Don't retain columns that are not valid direction vectors.
-		vecs.Clone(vecs.Slice(0, d, 0, n))
-	}
-	vars = svd.Values(nil)
-	var f float64
-	if weights == nil {
-		f = 1 / float64(n-1)
-	} else {
-		f = 1 / (floats.Sum(weights) - 1)
-	}
-	for i, v := range vars {
-		vars[i] = f * v * v
-	}
-	return vecs, vars, true
+	return c.ok
 }
 
-// CC is the result of a canonical correlation analysis. A CC should not be used
-// unless it has been returned by a successful call to CanonicalCorrelations.
+// Vectors returns the component direction vectors of a principal components
+// analysis. The vectors are returned in the columns of a d×min(n, d) matrix.
+// If dst is not nil it must either be zero-sized or be a d×min(n, d) matrix.
+// dst will  be used as the destination for the direction vector data. If dst
+// is nil, a new mat64.Dense is allocated for the destination.
+func (c *PC) Vectors(dst *mat64.Dense) *mat64.Dense {
+	if !c.ok {
+		panic("stat: use of unsuccessful principal components analysis")
+	}
+
+	if dst == nil {
+		dst = &mat64.Dense{}
+	} else if d, n := dst.Dims(); (n != 0 || d != 0) && (d != c.d || n != min(c.n, c.d)) {
+		panic(matrix.ErrShape)
+	}
+	dst.VFromSVD(c.svd)
+	return dst
+}
+
+// Vars returns the column variances of the principal component scores,
+// b * vecs, where b is a matrix with centered columns. Variances are returned
+// in descending order.
+// If dst is not nil it is used to store the variances and returned.
+// Vars will panic if the receiver has not successfully performed a principal
+// components analysis or dst is not nil and the length of dst is not min(n, d).
+func (c *PC) Vars(dst []float64) []float64 {
+	if !c.ok {
+		panic("stat: use of unsuccessful principal components analysis")
+	}
+	if dst != nil && len(dst) != min(c.n, c.d) {
+		panic("stat: length of slice does not match analysis")
+	}
+
+	dst = c.svd.Values(dst)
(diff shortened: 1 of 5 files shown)


pca_cca.go
```diff
@@ -13,51 +13,96 @@ import (
...
-// On successful completion, the principal component vectors are returned in the
-// columns of vecs as a d×min(n,d) matrix, and the column variances of the
-// principal component scores, b * vecs where b is a with centered columns, are
-// returned in vars as a min(n,d)-long slice in descending sort order.
-//
-// On failure, vecs and vars are returned nil and ok is returned false.
-func PrincipalComponents(a mat64.Matrix, weights []float64) (vecs *mat64.Dense, vars []float64, ok bool) {
-	n, d := a.Dims()
-	if weights != nil && len(weights) != n {
+// PrincipleCompoents returns true when the analysis is successful.
```
btracey: returns whether the analysis was successful.

pca_cca.go
```diff
@@ -13,51 +13,96 @@ import (
...
+		// We are less strict here because SVD may retain more data
+		// than we may want, which we then discard below. Resizing
+		// correctly between repeated calls may be onerous.
+		dst.Reset()
+	}
+	dst.VFromSVD(c.svd)
+	if c.n < c.d {
 		// Don't retain columns that are not valid direction vectors.
-		vecs.Clone(vecs.Slice(0, d, 0, n))
+		dst = dst.Slice(0, c.d, 0, c.n).(*mat64.Dense)
```
btracey: This doesn't have to change now, but I don't understand why we sometimes compute the full SVD when we only use the thin result.

pca_cca.go
```diff
@@ -13,51 +13,96 @@ import (
...
 	}
-	vars = svd.Values(nil)
+	return dst
+}
+
+// Vars returns the column variances of the principal component scores,
+// b * vecs where b is a with centered columns. Variances are returned
```
btracey: is the matrix a with

btracey:
Idea looks good to me.

pca_cca.go
kortschak: And spelling.

pca_cca.go
kortschak: That is outstanding. Thank you. Yes it does have to change now - I can now be strict on sizing.

I blame myself for this weirdness.

kortschak: It would be nice to have IsZero for this use - as you will see.

kortschak:
I have the user-facing allocations reduced, but there is potential for more saving in the internal slices and matrices in CC. This really wants to have pooling. It seems to me that the matrix pooling code could be put into internal and then shared. That would be another PR (obviously).

Also note that the changes here do not improve things really until https://github.com/gonum/matrix/pull/417 is merged.

PTAL

Also, @Armadilloa16, please take a look.

btracey:
We do a lot of temporary allocations in a lot of places for slices as well. If we move pooling into internal we could also use pooling for those allocations.

pca_cca.go
```diff
@@ -169,58 +152,134 @@ func (c CC) Right(spheredSpace bool) (vecs *mat64.Dense) {
...
+	} else if d, n := dst.Dims(); (n != 0 || d != 0) && (n != c.yd || d != c.xd) {
+		panic(matrix.ErrShape)
+	}
+	dst.UFromSVD(c.c)
+	if spheredSpace {
+		return dst
 	}
-	return matrix.SVDFull
+
+	var xv mat64.Dense
```
btracey: I'm assuming the algorithm is correct

kortschak:
> If we move pooling into internal we could also use pooling for those allocations.

That was my intention, yes.

pca_cca.go
```diff
@@ -136,7 +119,7 @@ func (c CC) Right(spheredSpace bool) (vecs *mat64.Dense) {
 //
 // Canonical correlation analysis finds associations between two sets of
 // variables on the same observations by finding linear combinations of the two
-// sphered datasets that maximise the correlation between them.
+// sphered datasets that maximize the correlation between them.
```
kortschak: I'm completely consistent here - internal comments are written in English my be, user-facing documentation is written in American.

The price of the benefits of cultural emperialism are what they are :)

pca_cca.go
```diff
@@ -13,51 +13,89 @@ import (
...
+	c.svd, c.ok = svdFactorizeCentered(c.svd, a, weights)
+	if c.ok {
+		c.weights = append(c.weights[:0], weights...)
+	}
+	return c.ok
+}
+
+// Vectors returns the component direction vectors of a principal components
+// analysis. The vectors are returned in the columns of a d×n matrix.
+// If dst is not nil it must either be zero-sized or be a d×n matrix. dst will
```
kortschak: I'll wait until others have weighed in on this then.

@vladimir-ch @Armadilloa16 ?

pca_cca.go
```diff
@@ -13,51 +13,89 @@ import (
...
-// On successful completion, the principal component vectors are returned in the
-// columns of vecs as a d×min(n,d) matrix, and the column variances of the
-// principal component scores, b * vecs where b is a with centered columns, are
-// returned in vars as a min(n,d)-long slice in descending sort order.
-//
-// On failure, vecs and vars are returned nil and ok is returned false.
-func PrincipalComponents(a mat64.Matrix, weights []float64) (vecs *mat64.Dense, vars []float64, ok bool) {
-	n, d := a.Dims()
-	if weights != nil && len(weights) != n {
+// PrincipleComponents returns whether the analysis was successful.
```
kortschak: Done

pca_cca.go
```diff
@@ -13,51 +13,89 @@ import (
...
+	ok      bool
+}
+
+// PrincipalComponents performs a weighted principal components analysis on the
+// matrix of the input data which is represented as an n×d matrix a where each
+// row is an observation and each column is a variable.
 //
 // PrincipalComponents centers the variables but does not scale the variance.
 //
 // The slice weights is used to weight the observations. If weights is nil, each
```
kortschak: Done

pca_cca.go
```diff
@@ -13,51 +13,89 @@ import (
...
+	return dst
+}
 
-	vecs = &mat64.Dense{}
-	vecs.VFromSVD(svd)
-	if n < d {
-		// Don't retain columns that are not valid direction vectors.
-		vecs.Clone(vecs.Slice(0, d, 0, n))
+// Vars returns the column variances of the principal component scores,
+// b * vecs where b is a matrix with centered columns. Variances are returned
```
kortschak: Done

pca_cca.go
```diff
@@ -13,51 +13,89 @@ import (
...
+// in descending sort order.
```
kortschak: Removed

pca_cca.go
```diff
@@ -13,51 +13,89 @@ import (
...
+// If dst is not nil is is used to store the variances and returned.
```
kortschak: Done

pca_cca.go
kortschak: I have not changed the algorithm - it went through an extensive PR.

pca_cca.go
kortschak: With the addition of a test for a wide matrix, I'm happy that this is the correct behaviour.

kortschak:
This is ready for a final review now.

PTAL

pca_cca.go
Armadilloa16: PCA is useful on fat matrices. The most common example would be for dimension reduction of data with more variables than observations. Unless I misunderstood `fat' in this context? Also I think there is already a test for a wide matrix, annotated `Truncated iris data transposed to check for operation on fat input.` ?

Armadilloa16: Also yeah as far as I can tell from the current implementation it should be `d x min(d,n)`, however it really should be `d x min(d, n-1)` because due to the centering done, the `n` th singular value will always be zero when `d < n`. I'm not sure how to implement this though other than just checking for it and removing the last column if necessary -- which seems ugly, I guess just leave the implementation as it is but a note in the documentation is probably appropriate. End of the day having an extra useless principal component vector in some cases won't do any harm I suppose.

kortschak: >Truncated iris data transposed to check for operation on fat input.

That was added in this PR, but yes, there was one previously.

kortschak: Fixed.

kortschak:
*ping*

Armadilloa16:
This LGTM apart from one bit that I don't completely understand (but apparently has to do with optimizing allocation), and I'll just trust you on that bit. :)

kortschak:
@Armadilloa16 which part don't you understand?

Armadilloa16:
Line 45 of pca_cca.go:

'
c.weights = append(c.weights[:0], weights...)
'

kortschak:
Ah, that truncates `c.weights` to zero length and then appends from there. This means that any allocation of `c.weights` is reused. It is possible to use this without preparation, since it is valid to slice a nil slice to zero length (it is still nil). You may want to read [this](https://blog.golang.org/go-slices-usage-and-internals) and [this](https://blog.golang.org/slices).

pca_cca.go
```diff
@@ -169,58 +153,134 @@ func (c CC) Right(spheredSpace bool) (vecs *mat64.Dense) {
...
+}
+
+// Corrs returns the canonical correlations, using dst if it is not nil.
+// If dst is not nil and len(dst) does not match the number of columns in
+// the y input matrix, Corrs will panic.
+func (c *CC) Corrs(dst []float64) []float64 {
+	if !c.ok {
+		panic("stat: canonical correlations missing or invalid")
+	}
+
```
btracey: remove \n?

kortschak: I'm OK with this here.

PR #152, opened 2016-12-19
Add Mahalanobis distance and update Normal.LogProb

btracey:


btracey:
Did not mean to add the KL files.

PR #153, opened 2016-12-19
Addmahal

btracey:


statmat_test.go
```diff
@@ -262,6 +262,36 @@ func TestCorrCov(t *testing.T) {
...
+					0.8, 0.3, 0.1,
+					0.3, 0.7, -0.1,
+					0.1, -0.1, 7}),
+			ans: 1.9251757377680914,
+		},
+	} {
+		var chol mat64.Cholesky
+		ok := chol.Factorize(test.Sigma)
+		if !ok {
+			panic("bad test")
```
sbinet: use `t.Fatal` instead?

kortschak: This is an internal panic, not a test panic.

btracey:
Ping?

btracey:
It is the same.

PR #157, opened 2017-02-09
add normal quantile

btracey:


distmv/normal.go
```diff
@@ -246,6 +246,29 @@ func (n *Normal) Prob(x []float64) float64 {
 	return math.Exp(n.LogProb(x))
 }
 
+// Quantile returns the multi-dimensional inverse cumulative distribution function.
+// len(x) must equal len(p), and if x is non-nil, len(x) must also equal len(p).
+// If x is nil, a new slice will be allocated and returned, otherwise the quantile
+// will be stored in-place into x. All of the values of p must be between 0 and 1,
```
kortschak: inclusive

distmv/normal.go
```diff
@@ -246,6 +246,29 @@ func (n *Normal) Prob(x []float64) float64 {
 	return math.Exp(n.LogProb(x))
 }
 
+// Quantile returns the multi-dimensional inverse cumulative distribution function.
+// len(x) must equal len(p), and if x is non-nil, len(x) must also equal len(p).
```
kortschak: I don't understand this.

Also, at the moment there is no explicit length check for `x` when `x != nil`; the panic is potentially a runtime bounds check if `len(x) < len(p)`, otherwise a panic in `TransformNormal`.

distmv/normal.go
```diff
@@ -246,6 +246,29 @@ func (n *Normal) Prob(x []float64) float64 {
...
+// If x is nil, a new slice will be allocated and returned, otherwise the quantile
+// will be stored in-place into x. All of the values of p must be between 0 and 1,
+// or Quantile will panic.
+func (n *Normal) Quantile(x, p []float64) []float64 {
+	dim := n.Dim()
+	if len(p) != dim {
+		panic(badSizeMismatch)
```
kortschak: `badInputLength`?

(I'm not sure that `badSizeMismatch` is a good label anyway - what does it mean? A really bad size mismatch or a size mismatch is not actually a size mismatch).

distmv/normal.go

distmv/normal.go
btracey: Added the explicit check for size.

kortschak: I still don't understand "len(x) must equal len(p), and if x is non-nil, len(x) must also equal len(p)." the first "len(x)" is presumably supposed to be something else, no?

distmv/normal.go
kortschak: What is `badSizeMismatch` supposed to be used for? I think we should bring these constants together - having them spread around in separate files the way they are leads to this kind of problem.

distmv/normal.go
btracey: Sorry, the comment was non-sensical. Fixed.

btracey:
ping

btracey:
Thanks

PR #159, opened 2017-02-17
Add distance functions between probability distributions

btracey:


btracey:
Two thoughts on this

1) I'm not sure how to test Wasserstein. The definition isn't one that's implementable as far as I understand. I can independently code the formula for Gaussians?

2) I wonder if these are better implemented as
````
type Bhattacharyya struct{}
func (Bhattacharyya) DistNormal(l,r *Normal) float64{blah}

// in other code
bc := distmv.Bhattacharyya{}.DistNormal(a,b)
````

Couple of advantages. One, it keeps all the similar functionality in the same place, so (many of them) can have a DistUniform, DistStudentsT, without needing a zillion top-level functions. Secondly, it makes using interfaces really easy

````
type DistNormaler interface {
DistNormal(a,b *distmv.Normal)
}
````

Without having to build wrapper types. I already have code that uses this interface.

File: distmv/statdist.go [hunk 1]
```diff
@@ -0,0 +1,163 @@
+// Copyright ©2016 The gonum Authors. All rights reserved.
```
btracey: In the past, we've had the copyright year as the year the code was written, and not the year the code was submitted. This was written last year, so should these years stay?

distmv/statdist.go
```diff
@@ -0,0 +1,163 @@
...
+
+	"github.com/gonum/floats"
+	"github.com/gonum/matrix/mat64"
+	"github.com/gonum/stat"
+)
+
+// TODO(btracey): Need to add dim checks
+
+// BhattacharyyaNormal computes the Bhattacharyya distance between l and r.
+// The dimension of the input distributions must match or the function will panic.
```
btracey: Or just in how I use it. Changed.

distmv/statdist.go
kortschak: OK, leave it as 2016. Thanks.

distmv/statdist.go
```diff
@@ -0,0 +1,189 @@
...
+type Wasserstein struct{}
+
+// DistNormal returns the Wasserstein distance between normal distributions l and r.
+// The dimensions of the input distributions must match or DistNormal will panic.
+//
+// The Wasserstein distance for Normal distributions is
+//  d^2 = ||m_l - m_r||_2^2 + Tr(Σ_l + Σ_r - 2(Σ_l^(1/2)*Σ_r*Σ_l^(1/2))^(1/2))
+// For more information, see
+//  http://djalil.chafai.net/blog/2010/04/30/wasserstein-distance-between-two-gaussians/
+func (Wasserstein) DistNormal(l, r *Normal) float64 {
```
kortschak: Any ideas on testing this yet?

btracey:
Shoot, sorry, I didn't see that you had approved this, and now there are a bunch of compilation issues. 

My plan is to start a new pull request with a fixed version of a subset of this.  It will contain the functions  I'm using for a paper that I'm about to submit. It needs #183 though. Is that okay @kortschak ?

kortschak:
The compilation errors are easily resolvable. Is there are reason not to just fix this?

btracey:
Sure, I can do that

btracey:
Done.

kortschak:
Still LGTM

PR #160, opened 2017-02-17
Fix incorrect LatinHypercube documentation and add code for Uniform c…

btracey:
…onvenience

### DIFF
--- a/distmv/general.go
+++ b/distmv/general.go
@@ -10,6 +10,7 @@ var (
 	badReceiver      = "distmv: input slice is not nil or the correct length"
 	badSizeMismatch  = "distmv: size mismatch"
 	badZeroDimension = "distmv: zero dimensional input"
+	nonPosDimension  = "distmv: non-positive dimension input"
 )
 
 const logTwoPi = 1.8378770664093454835606594728112352797227949472755668
--- a/distmv/uniform.go
+++ b/distmv/uniform.go
@@ -44,6 +44,25 @@ func NewUniform(bnds []Bound, src *rand.Source) *Uniform {
 	return u
 }
 
+// NewUnitUniform creates a new Uniform distribution over the dim-dimensional
+// unit hypercube. That is, a uniform distribution where each dimension has
+// Min = 0 and Max = 1.
+func NewUnitUniform(dim int, src *rand.Source) *Uniform {
+	if dim <= 0 {
+		panic(nonPosDimension)
+	}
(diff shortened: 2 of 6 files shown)


distmv/general.go
```diff
@@ -6,10 +6,11 @@
...
 var (
-	badQuantile      = "distmv: quantile not between 0 and 1"
-	badReceiver      = "distmv: input slice is not nil or the correct length"
-	badSizeMismatch  = "distmv: size mismatch"
-	badZeroDimension = "distmv: zero dimensional input"
+	badQuantile        = "distmv: quantile not between 0 and 1"
+	badReceiver        = "distmv: input slice is not nil or the correct length"
+	badSizeMismatch    = "distmv: size mismatch"
+	badZeroDimension   = "distmv: zero dimensional input"
+	badNonPosDimension = "distmv: non-positive dimension input"
```
kortschak: I think the prefix of bad is not helpful here. This is the same issue as `badSizeMismatch`. `nonPositiveDimension` seems more meaningful.

distmv/uniform.go
```diff
@@ -44,6 +44,22 @@ func NewUniform(bnds []Bound, src *rand.Source) *Uniform {
 	return u
 }
 
+func NewUnitUniform(dim int, src *rand.Source) *Uniform {
```
kortschak: Doc comment here please.

distuv/uniform.go
```diff
@@ -9,6 +9,8 @@ import (
 	"math/rand"
 )
 
+var UnitUniform = Uniform{Min: 0, Max: 1}
```
kortschak: Doc comment here please.

distmv/general.go
btracey: Sorry, I thought that we were using `bad` as a generic panic prefix. I guess I misunderstood the previous suggestion.

distmv/uniform.go
btracey: Done.

distuv/uniform.go
btracey: Done.

distmv/general.go
kortschak: "bad" just means that, that the thing was bad without necessarily saying why. So "badSizeMismatch" would either be "badSize" or preferably "sizeMismatch" (since that is the message). The same argument holds here.

btracey: Done.

PR #167, opened 2017-03-06
stat/distuv: add example for Normal

### LINKED ISSUE #166: stat/distuv: add an example for Normal
So far, there's no obvious usage example for, _eg_, `distuv.Normal`.

A simple example could go a long way for beginners.

sbinet:
Updates #166

### DIFF
--- a/distuv/norm_example_test.go
+++ b/distuv/norm_example_test.go
@@ -1,0 +1,35 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package distuv_test
+
+import (
+	"fmt"
+
+	"github.com/gonum/stat"
+	"github.com/gonum/stat/distuv"
+)
+
+func ExampleNormal() {
+	// Create a normal distribution
+	dist := distuv.Normal{
+		Mu:    2,
+		Sigma: 5,
+	}
+
+	data := make([]float64, 1e5)
+
+	// Draw some random values from the standard normal distribution
+	for i := range data {
+		data[i] = dist.Rand()
+	}
+
+	mean, std := stat.MeanStdDev(data, nil)
+	meanErr := stat.StdErr(std, float64(len(data)))
+
+	fmt.Printf("mean= %1.1f ± %0.1v\n", mean, meanErr)
+
+	// Output:
+	// mean= 2.0 ± 0.02
+}


sbinet:
@ebusato

distuv/norm_example_test.go
```diff
@@ -0,0 +1,43 @@
...
+	const (
+		nmax = 1000
+		seed = 1234
+	)
+
+	// Create a normal distribution
+	dist := distuv.Normal{
+		Mu:     0,
+		Sigma:  1,
+		Source: rand.New(rand.NewSource(seed)),
```
btracey: Can you have the main example be using `nil`? We don't want the typical user to blindly copy this code and then get the same random numbers many times. Using a non-nil source is an advanced use that should be done with care.

distuv/norm_example_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/gonum/stat/distuv"
+)
+
+func Example_normal() {
+	const (
+		nmax = 1000
+		seed = 1234
+	)
+
+	// Create a normal distribution
```
btracey: Maybe do a non-standard normal? We have a variable for standard normal it might be nice to highlight

btracey:
Can we say it updates 116? It would be nice to have an example that uses the CDF and log probabilities.

distuv/norm_example_test.go
sbinet: done.

distuv/norm_example_test.go
sbinet: done.

distuv/norm_example_test.go
sbinet: ah! but then travis won't be happy... (IIRC the seed isn't the same.)

sbinet:
revert to use an explicit `Source`. PTAL.

btracey:
Sorry, I thought my message had sent hours ago.

distuv/norm_example_test.go
kortschak: Please use a `nil` source and just don't have an expected output. We don't need this tested.

sbinet: changed it to a `nil` source, but test the mean and std-dev are within some tolerance instead.

sbinet:
PTAL

distuv/norm_example_test.go
```diff
@@ -0,0 +1,45 @@
...
+	data := make([]float64, nmax)
+
+	// Draw some random values from the standard normal distribution
+	for i := 0; i < nmax; i++ {
+		data[i] = dist.Rand()
+	}
+
+	mean, std := stat.MeanStdDev(data, nil)
+
+	const tol = 0.1
```
btracey: I think more educational is to:
- Increase samples to 100,000
- Actually compute the error in mean and std and print them

````
mean, std := stat.MeanStdDev(data, nil)
meanErr := stat.StdErr(std, float64(nmax))
stdErr := std * math.Sqrt(2/(float64(nmax)-1))

fmt.Printf("mean = %1.1f +/- %0.1v\n", mean, meanErr)
fmt.Printf("std-dev = %1.1f +/- %0.1v\n", std, stdErr)

// Output:
// mean=    2.0 +/- 0.02
// std-dev= 5.0 +/- 0.02
````

File: distuv/norm_example_test.go [hunk 4]
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"fmt"
+
+	"github.com/gonum/floats"
+	"github.com/gonum/stat"
+	"github.com/gonum/stat/distuv"
+)
+
+func ExampleNormal() {
+	const nmax = 1000
```
btracey: I would write it "nSamp", or "samp"

sbinet: removed completely.

distuv/norm_example_test.go
sbinet: done.

btracey:
LGTM

sbinet:
@kortschak : is the latest CL fine for you as well?

distuv/norm_example_test.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+func ExampleNormal() {
+	// Create a normal distribution
+	dist := distuv.Normal{
+		Mu:    2,
+		Sigma: 5,
+	}
+
+	data := make([]float64, 100000)
```
kortschak: s/100000/1e5/

distuv/norm_example_test.go
```diff
@@ -0,0 +1,39 @@
...
+	// Draw some random values from the standard normal distribution
+	for i := range data {
+		data[i] = dist.Rand()
+	}
+
+	mean, std := stat.MeanStdDev(data, nil)
+	meanErr := stat.StdErr(std, float64(len(data)))
+	stdErr := std * math.Sqrt(2/(float64(len(data))-1))
+
+	fmt.Printf("mean=    %1.1f +/- %0.1v\n", mean, meanErr)
```
kortschak: s|+/-|±|g

distuv/norm_example_test.go
```diff
@@ -0,0 +1,39 @@
...
+	data := make([]float64, 100000)
+
+	// Draw some random values from the standard normal distribution
+	for i := range data {
+		data[i] = dist.Rand()
+	}
+
+	mean, std := stat.MeanStdDev(data, nil)
+	meanErr := stat.StdErr(std, float64(len(data)))
+	stdErr := std * math.Sqrt(2/(float64(len(data))-1))
```
kortschak: Are you sure about this? Shouldn't it be `dist.Sigma * dist.Sigma * math.Sqrt(2/(float64(len(data))-1))`?

btracey: I think what you wrote is the standard error in the variance. The error in the standard deviation would be (roughly) square root that, so 

````
std * math.Sqrt(math.Sqrt(2/float64(len(data)-1))
````

I'm not sure what we should do about `std` vs. `dist.Sigma`. Mathematically, it's only accurate with `dist.Sigma`, though if you know `dist.Sigma` there's no point in computing an error in the standard deviation. Similarly for the mean, one does not normally know the standard deviation, and so one estimates it with data (as is done here now). We may want to make a comment to that effect though.

Comment (kortschak) [inline distuv/norm_example_test.go]: Does this actually help the example? It seems to me that the example should just be showing how to draw a sample, so it would be enough to

```
	mean, std := stat.MeanStdDev(data, nil)
	meanErr := stat.StdErr(std, float64(len(data)))
	
	fmt.Printf("mean=    %1.1f ± %0.1v\n", mean, meanErr)
```

btracey: SGTM

distuv/norm_example_test.go
sbinet: done.

distuv/norm_example_test.go
sbinet: done.

distuv/norm_example_test.go
sbinet: done

PR #168, opened 2017-03-07
Store sigma in Normal and StudentsT

btracey:
Currently, we throw sigma away, and recompute it if necessary. This PR keeps sigma. This fixes an issue with concurrent calling of methods. In addition, however, it removes any possible issues with reconstructing a badly-conditioned sigma from its Cholesky decomposition, and avoids an extra n^3 work if sigma does need to be recomputed. The complexity of the implementation and difficulties listed above is not worth the memory savings in some cases, especially since the memory of the type is already O(n^2)

### DIFF
--- a/distmv/normal.go
+++ b/distmv/normal.go
@@ -7,7 +7,6 @@ package distmv
 import (
 	"math"
 	"math/rand"
-	"sync"
 
 	"github.com/gonum/floats"
 	"github.com/gonum/matrix/mat64"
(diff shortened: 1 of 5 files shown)


btracey:
Fixes #163

btracey:
Not the right place for the question, but I've heard people say that and I don't understand why it's true. Reading and writing to a global variable concurrently is a race condition, and the implementation is allowed to assume no race conditions. Thus, it can tell the assignment to the global variable cannot be observed, and so it can optimize away all but the last iteration of the loop.

PR #170, opened 2017-03-18
stat: use the updated decomposition extraction API

kortschak:
I have left the CC methods `Left` and `Right` as they are rather than renaming to `LeftTo` and `RightTo`, which they probably should be changed to.

@btracey @vladimir-ch Please take a look.

Depends on https://github.com/gonum/matrix/pull/436.

### DIFF
--- a/pca_cca.go
+++ b/pca_cca.go
@@ -52,18 +52,17 @@ func (c *PC) PrincipalComponents(a mat64.Matrix, weights []float64) (ok bool) {
 // If dst is not nil it must either be zero-sized or be a d×min(n, d) matrix.
 // dst will  be used as the destination for the direction vector data. If dst
 // is nil, a new mat64.Dense is allocated for the destination.
-func (c *PC) Vectors(dst *mat64.Dense) *mat64.Dense {
+func (c *PC) VectorsTo(dst *mat64.Dense) *mat64.Dense {
 	if !c.ok {
 		panic("stat: use of unsuccessful principal components analysis")
 	}
 
-	if dst == nil {
-		dst = &mat64.Dense{}
-	} else if d, n := dst.Dims(); (n != 0 || d != 0) && (d != c.d || n != min(c.n, c.d)) {
-		panic(matrix.ErrShape)
+	if dst != nil {
+		if d, n := dst.Dims(); (n != 0 || d != 0) && (d != c.d || n != min(c.n, c.d)) {
+			panic(matrix.ErrShape)
+		}
 	}
-	dst.VFromSVD(c.svd)
-	return dst
+	return c.svd.VTo(dst)
 }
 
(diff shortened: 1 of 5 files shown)


pca_cca.go
```diff
@@ -57,13 +57,12 @@ func (c *PC) Vectors(dst *mat64.Dense) *mat64.Dense {
 		panic("stat: use of unsuccessful principal components analysis")
 	}
 
-	if dst == nil {
-		dst = &mat64.Dense{}
-	} else if d, n := dst.Dims(); (n != 0 || d != 0) && (d != c.d || n != min(c.n, c.d)) {
-		panic(matrix.ErrShape)
+	if dst != nil {
+		if d, n := dst.Dims(); (n != 0 || d != 0) && (d != c.d || n != min(c.n, c.d)) {
```
vladimir-ch: Split this into two lines?

pca_cca.go
```diff
@@ -228,23 +226,22 @@ func (c *CC) Left(dst *mat64.Dense, spheredSpace bool) *mat64.Dense {
 		panic("stat: canonical correlations missing or invalid")
 	}
 
-	if dst == nil {
-		dst = &mat64.Dense{}
-	} else if d, n := dst.Dims(); (n != 0 || d != 0) && (n != c.yd || d != c.xd) {
-		panic(matrix.ErrShape)
+	if dst != nil {
+		if d, n := dst.Dims(); (n != 0 || d != 0) && (n != c.yd || d != c.xd) {
```
vladimir-ch: Also here?

pca_cca.go
```diff
@@ -261,23 +258,22 @@ func (c *CC) Right(dst *mat64.Dense, spheredSpace bool) *mat64.Dense {
 		panic("stat: canonical correlations missing or invalid")
 	}
 
-	if dst == nil {
-		dst = &mat64.Dense{}
-	} else if d, n := dst.Dims(); (n != 0 || d != 0) && (n != c.yd || d != c.yd) {
-		panic(matrix.ErrShape)
+	if dst != nil {
+		if d, n := dst.Dims(); (n != 0 || d != 0) && (n != c.yd || d != c.yd) {
```
vladimir-ch: And here?

kortschak:
I left `Left` and `Right` etc, but think they probably should change. What do you think?

pca_cca.go
kortschak: :) This is an intentionally left sore to provoke the exporting of `IsZero` in mat64.

btracey:
Yea, I think they should change. Consistent naming is best for everyone.

kortschak:
This has been done in gonum/gonum.

kortschak:
Ooops. No it hasn't.

btracey:
It looks like this happened in the other repo.

Issue #172, opened 2017-03-27
stat/distuv: Adding F-distribution


dongweigogo:
Hello! I'm doing some statistical work with this package recently, and find it is very useful and practical. 

However the F-distribution (Fisher–Snedecor distribution) is still missing from the distuv module, that is required for the Levene test to assert equality of variances for Student's T test of independent samples. So, I think it is necessary to add the F-distribution to the type list. This will definitely make the package more powerful.

btracey:
This was done.

PR #173, opened 2017-04-15
stat/distuv: add F-distribution

zeroviscosity:
Addresses #172

### DIFF
--- a/distuv/f.go
+++ b/distuv/f.go
@@ -1,0 +1,118 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package distuv
+
+import (
+	"math"
+	"math/rand"
+
+	"github.com/gonum/mathext"
+)
+
+// F implements the F-distribution, a two-parameter continuous distribution
+// with support over the positive real numbers.
+//
+// The F-distribution has density function
+//  sqrt(((d1*x)^d1) * d2^d2 / ((d1*x+d2)^(d1+d2))) / (x * B(d1/2,d2/2))
+// where B is the beta function.
+//
+// For more information, see https://en.wikipedia.org/wiki/F-distribution
+type F struct {
+	D1     float64 // Degrees of freedom for the numerator
+	D2     float64 // Degrees of freedom for the denominator
+	Source *rand.Rand
+}
+
+// CDF computes the value of the cumulative density function at x.
+func (f F) CDF(x float64) float64 {
+	return mathext.RegIncBeta(f.D1/2, f.D2/2, f.D1*x/(f.D1*x+f.D2))
+}
+
+// ExKurtosis returns the excess kurtosis of the distribution.
+//
+// ExKurtosis returns NaN if the D2 parameter is less or equal to 8.
+func (f F) ExKurtosis() float64 {
+	if f.D2 <= 8 {
+		return math.NaN()
+	}
+	return (12 / (f.D2 - 6)) * ((5*f.D2-22)/(f.D2-8) + ((f.D2-4)/f.D1)*((f.D2-2)/(f.D2-8))*((f.D2-2)/(f.D1+f.D2-2)))
+}
+
+// LogProb computes the natural logarithm of the value of the probability
+// density function at x.
+func (f F) LogProb(x float64) float64 {
+	return 0.5*(f.D1*math.Log(f.D1*x)+f.D2*math.Log(f.D2)-(f.D1+f.D2)*math.Log(f.D1*x+f.D2)) - math.Log(x) - mathext.Lbeta(f.D1/2, f.D2/2)
+}
+
+// Mean returns the mean of the probability distribution.
+//
+// Mean returns NaN if the D2 parameter is less than or equal to 2.
+func (f F) Mean() float64 {
+	if f.D2 <= 2 {
+		return math.NaN()
+	}
+	return f.D2 / (f.D2 - 2)
+}
+
+// Mode returns the mode of the distribution.
+//
+// Mode returns NaN if the D1 parameter is less than or equal to 2.
+func (f F) Mode() float64 {
+	if f.D1 <= 2 {
+		return math.NaN()
+	}
+	return ((f.D1 - 2) / f.D1) * (f.D2 / (f.D2 + 2))
+}
+
+// NumParameters returns the number of parameters in the distribution.
+func (f F) NumParameters() int {
+	return 2
+}
+
+// Prob computes the value of the probability density function at x.
+func (f F) Prob(x float64) float64 {
+	return math.Exp(f.LogProb(x))
+}
+
+// Rand returns a random sample drawn from the distribution.
+func (f F) Rand() float64 {
+	u1 := ChiSquared{f.D1, f.Source}.Rand()
+	u2 := ChiSquared{f.D2, f.Source}.Rand()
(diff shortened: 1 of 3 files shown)


distuv/f.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// ExKurtosis returns the excess kurtosis of the distribution.
+//
+// ExKurtosis returns NaN if the D2 parameter is less or equal to 8.
+func (f F) ExKurtosis() float64 {
+	if f.D2 <= 8 {
+		return math.NaN()
+	}
+	num := 12 * (f.D1*(5*f.D2-22)*(f.D1+f.D2-2) + (f.D2-4)*math.Pow(f.D2-2, 2))
```
kortschak: `num := 12 * (f.D1*(5*f.D2-22)*(f.D1+f.D2-2) + (f.D2-4)*(f.D2-2)*(f.D2-2)`

distuv/f.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// Variance returns the variance of the probability distribution.
+//
+// Variance returns NaN if the D2 parameter is less than or equal to 4.
+func (f F) Variance() float64 {
+	if f.D2 <= 4 {
+		return math.NaN()
+	}
+	num := 2 * math.Pow(f.D2, 2) * (f.D1 + f.D2 - 2)
```
kortschak:
```
num := 2 * f.D2 * f.D2 * (f.D1 + f.D2 - 2)
den := f.D1 * (f.D2 - 2) * (f.D2 - 2) * (f.D2 - 4)
```

distuv/f_test.go
```diff
@@ -0,0 +1,86 @@
...
+		{13, 16, src},
+		{42, 31, src},
+		{77, 92, src},
+	} {
+		testF(t, b, i)
+	}
+}
+
+func testF(t *testing.T, f F, i int) {
+	tol := 1e-2
```
kortschak:
```
const (
	tol  = 1e-2
	n    = 2e6
	bins = 50
)
```

distuv/f.go
zeroviscosity: Updated.

distuv/f.go
zeroviscosity: Updated.

distuv/f_test.go
zeroviscosity: Updated.

distuv/f.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	num := 12 * (f.D1*(5*f.D2-22)*(f.D1+f.D2-2) + (f.D2-4)*(f.D2-2)*(f.D2-2))
+	den := f.D1 * (f.D2 - 6) * (f.D2 - 8) * (f.D1 + f.D2 - 2)
+	return num / den
+}
+
+// LogProb computes the natural logarithm of the value of the probability
+// density function at x.
+func (f F) LogProb(x float64) float64 {
+	return 0.5*(f.D1*math.Log(f.D1*x)+f.D2*math.Log(f.D2)-(f.D1+f.D2)*math.Log(f.D1*x+f.D2)) - math.Log(x*mathext.Beta(f.D1/2, f.D2/2))
```
btracey: `-math.Log(x) - mathext.LBeta(f.D1/2, f.D2/2)`

zeroviscosity: Good point. Updated.

distuv/f.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// ExKurtosis returns the excess kurtosis of the distribution.
+//
+// ExKurtosis returns NaN if the D2 parameter is less or equal to 8.
+func (f F) ExKurtosis() float64 {
+	if f.D2 <= 8 {
+		return math.NaN()
+	}
+	num := 12 * (f.D1*(5*f.D2-22)*(f.D1+f.D2-2) + (f.D2-4)*(f.D2-2)*(f.D2-2))
```
btracey: I don't know what typical parameters are of f distributions, but this looks badly conditioned to compute. I would split it up: first term is `((5*d2-22)/(d2-6)) * ((d1+d2-2)/(d2-8))` and similar for the second term. This should be much better behaved when d1 or d2 gets large, keeping each term O(1) rather than dividing two really large numbers.

distuv/f.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// Mode returns the mode of the distribution.
+//
+// Mode returns NaN if the D1 parameter is less than or equal to 2.
+func (f F) Mode() float64 {
+	if f.D1 <= 2 {
+		return math.NaN()
+	}
+	return (f.D1 - 2) * f.D2 / (f.D1 * (f.D2 + 2))
```
btracey: Same here, (f.D1-2)/(f.D1) * (f.D2)/(f.D2+2)

distuv/f.go
```diff
@@ -0,0 +1,120 @@
...
+func (f F) StdDev() float64 {
+	if f.D2 <= 4 {
+		return math.NaN()
+	}
+	return math.Sqrt(f.Variance())
+}
+
+// Variance returns the variance of the probability distribution.
+//
+// Variance returns NaN if the D2 parameter is less than or equal to 4.
```
btracey: NaN or Infinity?

distuv/f.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// CDF computes the value of the cumulative density function at x.
+func (f F) CDF(x float64) float64 {
+	return mathext.RegIncBeta(f.D1/2, f.D2/2, f.D1*x/(f.D1*x+f.D2))
+}
+
+// ExKurtosis returns the excess kurtosis of the distribution.
+//
+// ExKurtosis returns NaN if the D2 parameter is less or equal to 8.
```
btracey: NaN or Infinity?

distuv/f.go
zeroviscosity: I believe (and I hope someone will correct me if I'm wrong) that the parameters will be of the same magnitude as you'd see in the case of Chi-squared, so I don't think they'd get all that large. Nevertheless I completely agree that it makes sense to make it more robust to large values.

distuv/f.go
zeroviscosity: It approaches positive infinity as D2 approaches 4 from above, but I believe it's negative for values between 0 and 4. That said, scipy returns `inf` for 0 < D2 <= 4 and `nan` for D2 <= 0. Should it mimic that behaviour or just return Infinity?

distuv/f.go
zeroviscosity: Like the Variance, this approaches positive infinity as D2 approaches 8 from above. It's more complicated below 8, obviously, with some negative values and others that approach infinity. Unlike the Variance, though, scipy just returns `nan` if D2 <= 8.

distuv/f.go
zeroviscosity: Addressed.

distuv/f.go
zeroviscosity: Addressed.

distuv/f.go
btracey: I see. I assumed it was just unbounded for D2 < 4. If it's more complicated, then NaN is fine.

distuv/f.go
btracey: See above comment.

btracey:
If you could squash the commits, then we'll merge!

zeroviscosity:
Sounds good. All squashed.

PR #175, opened 2017-04-19
stat/combin: avoid overflow in Binomial

### LINKED ISSUE #174: A bug in combin.CombinationGenerator
I just found a bug in combin.CombinationGenerator. For n=30 and k=15, the generator produced nothing. The code is below:

    func main()
	        cg := combin.NewCombinationGenerator(30, 15)
	        com := make([]int, 15)
	        for i:=0;cg.Next();i++ {
	                com=cg.Combination(com)
	                fmt.Println(com)
	        }
    } 

So far, Same situation occurs when 
n=30 and k=14 (up to 16)
n=40 and k=14 (up to 26)
n=50 and k=20 (up to 33)
...
it seems that most cases are when k is close to half of n.

vladimir-ch:
PTAL @dongweigogo @btracey @kortschak 

Fixes #174

### DIFF
--- a/combin/combin.go
+++ b/combin/combin.go
@@ -35,15 +35,11 @@ func Binomial(n, k int) int {
 	if k > n/2 {
 		k = n - k
 	}
-	kfact := 1
-	for i := 2; i <= k; i++ {
-		kfact *= i
+	b := 1
+	for i := 1; i <= k; i++ {
+		b = (n - k + i) * b / i
 	}
-	num := 1
-	for i := n; i > n-k; i-- {
-		num *= i
-	}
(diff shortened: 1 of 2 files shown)


kortschak:
Not calculating the full factorial when it's not needed is pretty standard. I don't think this needs a comment. A test may be worthwhile.

vladimir-ch:
I added a test for the largest n for which the new code does not overflow. Could you please take a look again?

This issue raises a question whether it is worthwhile to have `Binomial` here when there is big.Int.Binomial and whether the whole package shouldn't use big.Int.

PR #178, opened 2017-04-21
stat/distuv: fix call parameters for SuffStat

### LINKED ISSUE #177: stat/distuv: incorrect length of suffStat parameter passed Normal.Fit
For normal distribution, the suffstat obviously has a length of 2, with mu and sigma. But in the distuv/norm.go source code, the length of suffStat in the Fit method is set to 1. That causes a panic.

    func (n *Normal) Fit(samples, weights []float64) {

	    suffStat := make([]float64, 1)                  //HERE

	    nSamples := n.SuffStat(samples, weights, suffStat)

	    n.ConjugateUpdate(suffStat, nSamples, []float64{0, 0})

    }

I think that should be set to 2.

kortschak:
Please take a look.

Fixes #177.

### DIFF
--- a/distuv/exponential.go
+++ b/distuv/exponential.go
@@ -71,9 +71,9 @@ func (Exponential) ExKurtosis() float64 {
 // If weights is nil, then all the weights are 1.
 // If weights is not nil, then the len(weights) must equal len(samples).
 func (e *Exponential) Fit(samples, weights []float64) {
-	suffStat := make([]float64, 1)
+	suffStat := make([]float64, e.NumSuffStat())
 	nSamples := e.SuffStat(samples, weights, suffStat)
-	e.ConjugateUpdate(suffStat, nSamples, []float64{0})
+	e.ConjugateUpdate(suffStat, nSamples, make([]float64, e.NumSuffStat()))
 }
 
 // LogProb computes the natural logarithm of the value of the probability density function at x.
@@ -104,6 +104,7 @@ func (Exponential) NumParameters() int {
 	return 1
 }
 
+// NumSuffStat returns the number of sufficient statistics for the distribution.
 func (Exponential) NumSuffStat() int {
 	return 1
 }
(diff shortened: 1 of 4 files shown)


distuv/norm_test.go
```diff
@@ -142,13 +141,23 @@ func TestNormalQuantile(t *testing.T) {
...
-		fmt.Println(math.Abs(got - ans[i]))
 		if !floats.EqualWithinAbsOrRel(got, ans[i], 1e-10, 1e-10) {
 			t.Errorf("Quantile mismatch. Case %d, want: %v, got: %v", i, ans[i], got)
 		}
 	}
 }
 
+func TestNormFitPanic(t *testing.T) {
+	n := Normal{Mu: 0, Sigma: 1}
+	func() {
```
vladimir-ch: Missing defer.

distuv/exponential_test.go
```diff
@@ -58,3 +58,14 @@ func TestExponentialScore(t *testing.T) {
 		testDerivParam(t, test)
 	}
 }
+
+func TestExponentialFitPanic(t *testing.T) {
+	e := Exponential{Rate: 2}
+	func() {
```
vladimir-ch: Missing defer.

distuv/exponential.go
```diff
@@ -199,13 +200,13 @@ func (e Exponential) StdDev() float64 {
 // of the samples.
 //
 // If weights is nil, the weights are assumed to be 1, otherwise panics if
-// len(samples) != len(weights). Panics if len(suffStat) != 1.
+// len(samples) != len(weights). Panics if len(suffStat) != NumSuffStat().
 func (Exponential) SuffStat(samples, weights, suffStat []float64) (nSamples float64) {
```
vladimir-ch: A side note in case we want to stabilize the API one day: the destination slice is usually the first.

distuv/norm_test.go
kortschak: Thanks. Fixed.

distuv/exponential_test.go
kortschak: Fixed.

distuv/exponential.go
kortschak: Yeah. I think, in line with @btracey's request that I hold off on the factorisation matrix extraction revisions, that should probably wait until the monorepoisation has happened.

There should probably be an issue in gonum/gonum for that to aggregate issues https://github.com/gonum/gonum/issues/4.

PR #181, opened 2017-05-02
stat/distmv: add Entropy to Uniform distribution

btracey:


distmv/uniform.go
```diff
@@ -97,6 +97,16 @@ func (u *Uniform) Dim() int {
 	return u.dim
 }
 
+// Entropy returns the differential entropy of the distribution.
+func (u *Uniform) Entropy() float64 {
+	// Entropy is log of the volume
```
kortschak: Period.

PR #185, opened 2017-05-06
stat/distmv: Add distance measures for Uniform, and fix Uniform source

btracey:


btracey:
I use these distances and interfaces in a paper I am working on.

distmv/statdist.go
```diff
@@ -12,6 +12,14 @@ import (
 	"github.com/gonum/stat"
 )
 
+type DistNormaler interface {
```
kortschak: I don't think these are justifiable here. Can you not define them in your package?

(Also, no docs).

btracey: I can define them in my package, but here seemed like a natural place to define them. Why aren't they justifiable?

kortschak: We don't use them anywhere and they are very small. Remember that we culled a large number of interfaces from mat64 for these reasons.

btracey: (playing devil's advocate to understand policy)

With `mat64` (in my perception), it was not clear that the interfaces were useful and sufficient. There's a similar problem with `distxv`, where there are tons of possible interfaces, and it's not fully clear which of them are powerful enough to merit inclusion. Here seems different. The types are clearly designed to a certain interface, and any code that wants to abstract over a distance metric will want to use this interface. Having the interface in `distmv`, rather than my package, allows different codes to communicate through the same interface instead of needing wrapper types (https://github.com/golang/go/issues/8082)

kortschak: 8082 is a problem, I agree, but the general principle is to define interfaces where they will be used (or at least something that could be described as the closest common dependency lending from evolutionary terminology). At the moment your packages are the only dependents. I would say that if other packages show a need for these interfaces because of use like what is described in 8082 (rather than just general use that doesn't cause syntactic mismatch) then these could be added, but at the moment it seems premature.

Possibly naively, I would render your use here (from the definitions of the interface) as a closer or method value.

btracey: SGTM. Interfaces removed

btracey:
Ping?

distmv/statdist_test.go
```diff
@@ -38,24 +38,64 @@ func TestBhattacharyyaNormal(t *testing.T) {
...
+		// Bhattacharyya should by symmetric
+		got2 := Bhattacharyya{}.DistUniform(b, a)
+		if math.Abs(got-got2) > 1e-14 {
+			t.Errorf("Bhattacharyya distance not symmetric")
 		}
 	}
 }
 
+// bhattacharryaSample finds an estimate of the Bhattacharrya coefficient through
+// sampling.
```
vladimir-ch: nit: double y, not double r

distmv/statdist_test.go
```diff
@@ -164,18 +204,46 @@ func TestKullbackLieblerNormal(t *testing.T) {
...
+			b:       NewUniform([]Bound{{-4, 1}, {-7, 10}}, rnd),
+			samples: 100000,
+			tol:     1e-2,
+		},
+	} {
+		a, b := test.a, test.b
+		want := klSample(a.Dim(), test.samples, a, b)
+		got := KullbackLeibler{}.DistUniform(a, b)
+		if math.Abs(want-got) > test.tol {
+			t.Errorf("Bhattacharyya mismatch, case %d: got %v, want %v", cas, got, want)
```
vladimir-ch: Wrong distance name.

distmv/statdist.go
```diff
@@ -139,6 +173,40 @@ func (KullbackLeibler) DistNormal(l, r *Normal) float64 {
...
+// l and r. The dimensions of the input distributions must match or DistUniform
+// will panic.
+func (KullbackLeibler) DistUniform(l, r *Uniform) float64 {
+	bl := l.Bounds(nil)
+	br := r.Bounds(nil)
+	if len(bl) != len(br) {
+		panic(badSizeMismatch)
+	}
+
+	// The KL is ∞ if x1 is not completely contained within x2, because then
```
vladimir-ch: What are `x1`, `x2`, `x`?

distmv/statdist_test.go
```diff
@@ -38,24 +38,64 @@ func TestBhattacharyyaNormal(t *testing.T) {
...
+
+func TestBhattacharyyaUniform(t *testing.T) {
+	rnd := rand.New(rand.NewSource(1))
+	for cas, test := range []struct {
+		a, b    *Uniform
+		samples int
+		tol     float64
+	}{
+		{
+			a:       NewUniform([]Bound{{-3, 2}, {-5, 8}}, rnd),
```
vladimir-ch: Test also non-overlapping case?

distmv/statdist_test.go
```diff
@@ -164,18 +204,46 @@ func TestKullbackLieblerNormal(t *testing.T) {
...
+
+func TestKullbackLieblerUniform(t *testing.T) {
+	rnd := rand.New(rand.NewSource(1))
+	for cas, test := range []struct {
+		a, b    *Uniform
+		samples int
+		tol     float64
+	}{
+		{
+			a:       NewUniform([]Bound{{-5, 2}, {-7, 12}}, rnd),
```
vladimir-ch: Add a non-overlapping test case?

distmv/statdist_test.go
```diff
@@ -164,18 +204,46 @@ func TestKullbackLieblerNormal(t *testing.T) {
...
+			t.Errorf("Case %d, KL mismatch: got %v, want %v", cas, got, want)
 		}
-		klmc /= float64(test.samples)
-		kl := KullbackLeibler{}.DistNormal(a, b)
-		if !floats.EqualWithinAbsOrRel(kl, klmc, test.tol, test.tol) {
-			t.Errorf("Case %d, KL mismatch: got %v, want %v", cas, kl, klmc)
+	}
+}
+
+func TestKullbackLieblerUniform(t *testing.T) {
```
vladimir-ch: Leibler

Also in TestKullbackLieblerNormal but it's out of the diff.

distmv/statdist_test.go
```diff
@@ -164,18 +204,46 @@ func TestKullbackLieblerNormal(t *testing.T) {
...
+		a, b := test.a, test.b
+		want := klSample(a.Dim(), test.samples, a, b)
+		got := KullbackLeibler{}.DistUniform(a, b)
+		if math.Abs(want-got) > test.tol {
+			t.Errorf("Bhattacharyya mismatch, case %d: got %v, want %v", cas, got, want)
 		}
 	}
 }
+
+// klSample finds an estimate of the Kullback-Liebler Divergence through sampling.
```
vladimir-ch: Leibler

distmv/statdist.go
btracey: Fixed.

distmv/statdist_test.go
btracey: Done.

distmv/statdist_test.go
btracey: Ah shoot.

distmv/statdist_test.go
btracey: Ah shoot

distmv/statdist_test.go
btracey: Done.

distmv/statdist_test.go
btracey: Fixed.

distmv/statdist_test.go
btracey: Blarg

distmv/statdist_test.go
```diff
@@ -38,24 +38,70 @@ func TestBhattacharyyaNormal(t *testing.T) {
...
+		}
+		// Bhattacharyya should by symmetric
+		got2 := Bhattacharyya{}.DistUniform(b, a)
+		if math.Abs(got-got2) > 1e-14 {
+			t.Errorf("Bhattacharyya distance not symmetric")
 		}
 	}
 }
 
+// bhattacharyyaSample finds an estimate of the Bhattacharrya coefficient through
```
kortschak: Bhattacharyya

distmv/statdist.go
```diff
@@ -139,6 +173,40 @@ func (KullbackLeibler) DistNormal(l, r *Normal) float64 {
...
+	br := r.Bounds(nil)
+	if len(bl) != len(br) {
+		panic(badSizeMismatch)
+	}
+
+	// The KL is ∞ if l is not completely contained within r, because then
+	// r(x) is zero when l(x) is non-zero for some x.
+	contained := true
+	for i, v := range bl {
+		if br[i].Min > v.Min || br[i].Max < v.Max {
```
kortschak: `v.Min < br[i].Min || br[i].Max < v.Max`

btracey:
Thank you

PR #188, opened 2017-07-21
Add triangle distribution

bpgray:
This adds a triangle distribution to the package, which is useful for business simulations when there is limited sample data available about input variables.


btracey:
Sorry, we aren't accepting pull requests on this package anymore. Could you have the pull request be in github.com/gonum/gonum instead?

btracey:
Also, we'll need a set of tests with it, you can look at some of the other distributions for example tests. Thanks for the PR!

bpgray:
Thank you for the quick reply. I'll submit a pull request at github.com/gonum/gonum and include a set of tests.

